PR #39, opened 2024-01-20
Don't Animate to initialPage

jezell:
Currently, the PdfViewer animates to the initialPage, but this causes undesirable flicker when the document is loaded. This change changes the initialPage loading to be instant.


espresso3389 (approved):
LGTM

Issue #4, opened 2023-12-13
 Is text selection possible on PDF?


mlican:
I'm interested in using the [pdfrx] for my Android application. I would like to know if it supports text selection functionality. Specifically, I would like to implement a feature where users can long-press on a PDF document and select/highlight text.

I've explored the documentation and codebase, but I couldn't find any specific information regarding text selection. Could you please clarify if this feature is supported or provide any guidance on how to achieve it?

Thank you for your assistance!

espresso3389:
I'm still working on that but on GitHub, you can see the "initial" and "draft" design of the underlying text extraction API and it's usage example (not yet finished);

- [PdfPageText](https://github.com/espresso3389/pdfrx/blob/master/lib/src/pdfrx_api.dart#L221)
- [example usage](https://github.com/espresso3389/pdfrx/blob/master/example/lib/main.dart#L45)

Basically, underlying API is almost freezed (not to be changed) but I should integrate text selection GUI code into PdfViewer widget anyway (I'm just working and I will take 1 week to finish I guess).

Task lists so far:

- On Desktop, if I use [SelectionArea](https://api.flutter.dev/flutter/material/SelectionArea-class.html) widget for text selection purpose, it blocks pan-to-scroll
- Search popup (how to customize or such)
- Hyperlink handling

Anyway, I'm still open to any opnions and suggestions so please feel free to ask anything to me!

espresso3389:
Oh, you can see the current progress on the demo site: https://espresso3389.github.io/pdfrx/

mlican:
> [0.2.1](https://pub.dev/packages/pdfrx/versions/0.2.1) is a new release that stabilizes the API;
> 
> * example code has possible demonstration code to select texts

Inaccurate Cursor Position and Selector Shaking Issue during PDF Zoom on Android.

I have noticed an issue while zooming in on PDF documents on Android. The cursor position for text selection seems to be inaccurate, and there is also a shaking phenomenon observed with the selector. This affects the precision and usability of selecting text within the PDF viewer. It would be greatly appreciated if this issue could be addressed and resolved. Thank you!

espresso3389:
I understood that. I'm still working on it.

Aruljebaraj:
Badly waiting for this feature

espresso3389:
Still work in progress; but I'm struggling with Flutter's text selection mechanism...

[...]

PR #40, opened 2024-01-20
Add Document rotation

wenchen:
I added a rotation parameter to the current code to rotate the page. When the page is rotated, the document is reloaded to draw the screen again. This ensures that the correct document size is displayed after the page is rotated.

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -31,6 +31,7 @@ class _MainPageState extends State<MainPage> {
   final controller = PdfViewerController();
   bool showSearchToolbar = false;
   final outline = ValueNotifier<List<PdfOutlineNode>?>(null);
+  int rotation = 0;
 
   @override
   void dispose() {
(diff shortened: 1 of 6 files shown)


espresso3389:
Please understand the code structure and please keep such UI related code change on widget codes only.

espresso3389:
By the way, why the rotation is document level parameter?
For what purpose do you want to do such thing?

espresso3389:
Another head-ache with the page rotation is text selection (#4) that I've been working on...

wenchen:
I tried several methods of page rotation, but it wasn't until I saw the demo page provided by pdf.js that I realized that it rotated the entire document after rotation. After some investigation, I found that I had to tell it the current rotation angle when adding a page. The PDFium test also seems to be the same. I think it should be possible to rotate a part of the page. I will look into how to handle it better. Thank you for your reply.

espresso3389:
Page rotation may be easily implemented on widget side.

So, in my opinion, such rotation can be done within PdfViewer widget. No lower layer should handle such UI side thing.

If I implement the rotation, I will do:

- Add document level (or page level) rotation angle in PdfViewerParams
- Update layout logic in _PdfViewerState._layoutPages
- Update drawing logic in _PdfViewerState._customPaint

Issue #36, opened 2024-01-19
PdfDocumentReference is disposed too aggressively / PdfViewer should support viewing a PdfDocument.


jezell:
If you have a PdfDocument in memory, there is not a good way to use it with the PdfViewer. 

PdfViewer should support direct rendering a PdfDocument rather than forcing the use of a PdfDocumentRef + PdfStore.

You can use the PdfStore.load method to return a PdfDocument as a PdfDocumentRef, but the magic dispose of PdfDocumentRef will get in the way and prevent you from reusing the PdfDocument:

```dart
  @override
  void removeListener(VoidCallback listener) {
    _listeners.remove(listener);
    if (_listeners.isEmpty) {
      dispose();
    }
  }
```

As soon as the viewer disposes, the reference will dispose itself, breaking the PdfDocument. Dispose should not be automatically called by the removeListener call, the reference should only be disposed when dispose is called manually. If you use PdfStore.load to try to manage the lifecycle of the document on your own, this forces you to add a dummy listener just to prevent it from being disposed if you need to reuse the reference later.

espresso3389:
#37 is anyway fixed.

Basically, PdfDocumentRef is introduced to keep the "same" instance between password trials (without valid password, PdfDocument cannot be obtained). If we do that without such mechanism, it's very difficult to deal with async PdfDocument creation process with password trials.

Anyaway, for your purpose, I just added PdfViewer.documentRef that directly accepts PdfDocumentRef.

espresso3389:
PdfViewer.documentRef will be removed on the next release. I realized that the primary constructor has the completely same signature to it :(

PR #42, opened 2024-01-22
Improve PDF loading and disposing by replacing PdfDocumentStore with PdfDocumentProvider

jezell:
As mentioned in #36, the current model makes it difficult to render a PdfDocument. In our case, this made it hard to move to pdfrx since all our code already uses PdfDocument. 

The current PdfStore loading model also has concurrency issues which can lead to downloading the same document multiple times if rebuilds happen quickly and also puts the viewer into an invalid state during rebuilds, which we immediately encountered with PdfViewer.uri.

This pull request takes some inspiration from NetworkImage and ImageProviders in the flutter framework to clean up the loading of documents, allow better detection of when the document should be reloaded, solve these concurrency issues and allowing for direct viewing of PdfDocument while retaining the PdfViewer.asset, PdfViewer.data, PdfViewer.file, PdfViewer.uri constructors.

PdfStore is replaced by PdfDocumentProvider which allows use of the == operator to better know when documents need to be reloaded due to widget changes, allows for automatic disposing of documents loaded by widgets, but also adds support for managing and disposing PdfDocument references without using a central store.


espresso3389:
The latest release is based on your PR but it changes several things:

- Add PdfDocumentProvider.uri, PdfDocumentProvider.file, PdfDocumentProvider.asset, ...
  - It's not easy to find classes that inherits PdfDocumentProvider
- PdfDocumentProvider.custom is not abstract; it's usage is fixed
- autoDispose is true by default in any case
   - if the user don't want to dispose it, they should set autoDispose=false explicity

espresso3389:
@jezell 
One question, why passwordProvider is moved to getDocument?
If the user opens same document multiple times, he is required to enter password again (without doing some extra work).

espresso3389:
I'm so sorry but I finally decided to remove codes based on your PdfDocumentProvider.

Now, PdfDocumentRef does almost everything which was provided by PdfDocumentProvider.
PdfDocumentRef.load has the actual PdfDocument loading code and it is protected by synchronized function now.

espresso3389:
My first concern is that with your changes, it no longer shares same PdfDocument between widgets. Every widget state internally creates its own instance. But if PdfDocumentRef loads PdfDocument, technically we can share the document instance though it is not supported by normal PdfViewer.uri or such.

jezell:
> One question, why passwordProvider is moved to getDocument?

It was not in the provider so that the provider object could be const. Function pointers are problematic with const objects because usually they can't really be passed a const reference. A better change would remove the callback entirely and clean up PdfDocumentReference a bit, but I was trying to make the minimal amount of changes with the commit.

> I'm so sorry but I finally decided to remove codes based on your PdfDocumentProvider.
> 
> Now, PdfDocumentRef does almost everything which was provided by PdfDocumentProvider. PdfDocumentRef.load has the actual PdfDocument loading code and it is protected by synchronized function now.

As long as the same logic is there doesn't matter too much. If this is how it is going to work, I think the autoDispose flag needs to be in the PdfViewer constructor that accepts the document. Our app started crashing again because the viewer is disposing the document we pass it. I'd think it would be better not to dispose it by default since if the caller loaded their own PdfDocument they should be responsible for disposing it.

The advantage of the providers over the current implementation is:

1) the == operator is a little cleaner and quicker than string comparisons.
2) Widget classes should always be const, the non provider model as implemented now can't be const because PdfDocumentReference is state that is stored on the widget instead of the state. This doesn't make a lot of sense because PdfReference is state and should be created and disposed by the State<PdfViewer>. The providers moved all the non const fields out of the widget to allow for use of const, which can be important for performance as it leads to a lot of optimizations in flutter and allowed the state to properly move to State<PdfViewer>. The primary thing that bugs me about the current model is that the state is being improperly managed and stored inside the Widget instead of inside the State. I think if the providers aren't going to be used, at a minimum the PdfReference creation should move inside the State<PdfViewer>, and the PdfDocumentReference should only be recreated if didUpdateWidget detects a different source being passed to the widget. Could definitely be done without introducing providers, but that was the real reason for the change. The providers were just a fairly clean way to do it that more closely matches how the Flutter SDK handles images.

espresso3389:
Mmm, yes, I understand what you're talking about.
But I want to keep authentication result in the PdfDocumentRef side somehow...
So should I move non-const states outside the PdfDocumentRef? I mean, using some global states like the previous implementation using PdfDocumentStore.

jezell:
> Mmm, yes, I understand what you're talking about. But I want to keep authentication result in the PdfDocumentRef side somehow... So should I move non-const states outside the PdfDocumentRef? I mean, using some global states like the previous implementation using PdfDocumentStore.

I think it's ok for the state to be inside the PdfReference like it is now, it's just not ok for the PdfReference to created inside the PdfViewer instead of the State<PdfViewer>. In order to move the reference to the state, you need the non const stuff that describes the source in another object. If the PdfViewer had the source properties itself like Uri? uriSource, String? fileSource, etc, then the state could create the reference. That's basically what the providers were accomplishing. I probably did a bad job of explaining this.

espresso3389:
My requirements are;

- Share the same PdfDocument between widgets
- Authentication trials should be done once

How should I do with the provider logic?

espresso3389:
Basically, why I want to introduce PdfDocumentRef instead of PdfDocument is that that encapsulates all the document loading and authentication process within it. If I can do that without changing PdfViewer constructors signature, I can accept any recommendations.

jezell:
> My requirements are;
> 
> * Share the same PdfDocument between widgets
> * Authentication trials should be done once
> 
> How should I do with the provider logic?

I think we have similar requirements because we also need to share PdfDocument between widgets. That's what we were using the RawPdfDocumentProvider for since we already had our own caching layer for the documents, but it I think the providers also make this easier if you want to have some caching built in, as it would allow for the provider to be the key for the cache. Something like this would work with the providers. I'd be happy to put together a new PR with caching supported if it helps:

```dart

class PdfDocumentCache {

    Map<PdfDocumentProvider, PdfDocumentReference> _documents = {};
  
    PdfDocumentReference getDocument(PdfDocumentProvider provider, PdfPasswordProvider? passwordProvider, bool firstAttemptByEmptyPassword) {
       if(!_documents.containsKey(provider)) {
           _documents[provider] = provider.getDocument(passwordProvider, firstAttemptByEmptyPassword);
       }
        return _documents[provider];
    }
    
}

class PdfDocumentCacheProvider extends InheritedWidget {
    PdfDocumentCacheProvider({ required this.cache, super.key });
    final PdfDocumentCache cache;
    
    // ...
}

class _PdfWidgetState extends State<PdfWidget> {

    void _widgetUpdated(PdfViewer? oldWidget) {
      // ...
      final cache = PdfDocumentCacheProvider.maybeOf(context);
      documentRef =  cache == null ? widget.provider.getDocument(
          widget.passwordProvider, widget.firstAttemptByEmptyPassword) : cache.getDocument(widget.provider, widget.passwordProvider, widget.firstAttemptByEmptyPassword);
      // ...
    }
}

```

jezell:
> Basically, why I want to introduce PdfDocumentRef instead of PdfDocument is that that encapsulates all the document loading and authentication process within it. If I can do that without changing PdfViewer constructors signature, I can accept any recommendations.

I don't think the PdfDocumentReference is a bad idea. It makes sense to have an object that encapsulates the state of the download, allows listening to progress, etc. which PdfDocument doesn't do (for example, in the ImageProvider from the Flutter SDK, a ImageStreamCompleter is returned, which looks a lot like PdfDocumentReference). The name could maybe be better like PdfDocumentCompleter, but I think the general idea is correct. The concern is more around where and how the PdfDocumentReference is being instantiated and making sure it moves inside the stateful part of the widget.

espresso3389:
Woops, 0343f81a4bf5792d91a9ffeb408a28bc52c0e02d fixes the error anyway.

espresso3389:
[0.4.21](https://pub.dev/packages/pdfrx/versions/0.4.21) is just released.
If you have any opinions, please let me know.

PR #94, opened 2024-03-04
fix Execution failed for task ':pdfrx:mergeReleaseNativeLibs'.

q200892907:
fix Execution failed for task ':pdfrx:mergeReleaseNativeLibs'.


espresso3389:
Basically, I think it's a problem of the app and not of pdfrx. So why pdfrx should do that?
Could you please explain more about the change?

PR #122, opened 2024-04-09
Fix RSS exploding when zoom in on large pages

yhyh0:
Might also worth considering that we crop the scaled image to cover only the visible rect... Any thoughts? Would that break too many parts?

Now it takes about 120mb for a 3000*4000 page to scale up to 2.0, it won't work well if we scale it all the way up to the max scale of 7 or higher

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,12 @@
+# 1.0.52
+
+- Fixes memory consumption control issues (Related: #121)
+
 # 1.0.51
 
-- FIXED: memory leak on _PdfPageViewState (#110)
+- FIXED: memory leak on \_PdfPageViewState (#110)
 - Remove dependency on dart:js_util (#109)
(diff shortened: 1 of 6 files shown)


yhyh0:
@espresso3389 closed the pr since I saw your changes, but the question remains.. Still value your thoughts when you get a chance to look at it.
And I notice a few blink(no pdf image showing up for a ms) when move/scale using the latest 1.0.52. I'm looking at it, but if you already know/fix the issue, please let me know :)

yhyh0:
@espresso3389 closed the pr since I saw your changes, but the question remains.. Still value your thoughts when you get a chance to look at it.
And I notice a few blink(no pdf image showing up for a ms) when move/scale using the latest 1.0.52. I'm looking at it, but if you already know/fix the issue, please let me know :)

espresso3389:
Yes, I also realized there are blinks on certain operation. I'm checking it.

PR #125, opened 2024-04-11
Reduce memory consumption by unloading page/document objects

yhyh0:
This PR is not ready to be merged. The work is not complete, and might not be the final solution.
It does provide some promising result though.

In a situation where a 3k x 4k page is being zoom to a scale of 6(with the partial image commit).
This is before:
截屏2024-04-11 10 44 06

And this is after:
截屏2024-04-11 10 48 36

Thoughts?

I should probably also try closing the doc early.

### DIFF
--- a/lib/src/pdfium/pdfrx_pdfium.dart
+++ b/lib/src/pdfium/pdfrx_pdfium.dart
@@ -397,6 +397,8 @@ class PdfDocumentPdfium extends PdfDocument {
         rotation: PdfPageRotation.values[pageData.rotation],
         page: pdfium_bindings.FPDF_PAGE.fromAddress(pageData.page),
       ));
+      pdfium.FPDF_ClosePage(
+          pdfium_bindings.FPDF_PAGE.fromAddress(pageData.page));
     }
     pdfDoc.pages = List.unmodifiable(pages);
     return pdfDoc;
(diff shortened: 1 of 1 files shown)


espresso3389:
I think "close" should be optional regardless of whether it is the default or not.

We should have some kind of flag named "keepMemoryCompact" or such to optimized for low memory consumption but user can disable it for performance even though I don't know how much we can get the performance gain if keep the cache...

espresso3389:
So, in the code we should have some functions like

- `ensurePageAvailable` or such to open the page handle before doing some task
- `releasePageIfNeeded` to close the page if the low memory option is enabled

espresso3389:
@yhyh0 Simple question. What kind of PDF document are you using; how many pages, how much size does it have. For general PDF files, loadDoc does not take so long.

So, my feeling is that your strategy seems better then the current structure. And I'm also interested in purging PdfDocument from the memory.
My suggestion is that we should measure the first document loading time. And if the loading time is less than some threshold, we uses close-everytime strategy; if not, we use the current keep-document-open strategy.

espresso3389:
And, anyway, could you please revert the documentRef changes? I want to merge the page unloading code at first. And for the document, we will do it after the merge.

espresso3389:
Anyway, PdfDocumentRef structure is not simple. It has password dialog interaction part and if we re-open frequently without user interaction, we should introduce some password cache mechanism at first (at least). Or we should re-architecture the things to handle the things better...

espresso3389:
I've fixed several isolate boundary errors on your code.

espresso3389:
[1.0.55](https://pub.dev/packages/pdfrx/versions/1.0.55) that contains the PR is released!

PR #126, opened 2024-04-12
Speed up the first render/Give a chance to cache thumbnail

yhyh0:
- [Skip rendering the first page if an initialPageNumber specified](https://github.com/espresso3389/pdfrx/commit/8048a50787ac6bae68abec984f3a7993065cdc0a)
- [Call onViewerReady only when the first page rendered; Speed up the first render by only rendering the necessary pages](https://github.com/espresso3389/pdfrx/commit/c34e71b304c3c7174a1f278ab39573e7d7e8ac16) 

The changes on ```onViewerReady``` gives a chance to show a loading widget and to save the first page image as a cached thumbnail(no need to load the doc if need a thumbnail later).
I thought adding a new ```onFirstPageRendered```, but  it might just makes more sense to postpone ```onViewerReady``` a bit waiting for the first render to make the 'ready' more ready.

### DIFF
--- a/lib/src/widgets/pdf_viewer.dart
+++ b/lib/src/widgets/pdf_viewer.dart
@@ -976,7 +976,7 @@ class _PdfViewerState extends State<PdfViewer>
                 widget.params.onViewerReady?.call(
                   _document!,
                   _controller!,
-                  newImage
+                  newImage.image
                 );
               }
             }
--- a/lib/src/widgets/pdf_viewer.dart
+++ b/lib/src/widgets/pdf_viewer.dart
@@ -207,6 +207,7 @@ class _PdfViewerState extends State<PdfViewer>
   double _minScale = 0.1;
   int? _pageNumber;
   bool _initialized = false;
+  bool _firstPageRendered = false;
   final List<double> _zoomStops = [1.0];
 
   final _pageImages = <int, ({ui.Image image, double scale})>{};
@@ -388,10 +389,6 @@ class _PdfViewerState extends State<PdfViewer>
                 widget.initialPageNumber;
             await _goToPage(
                 pageNumber: initialPageNumber, duration: Duration.zero);
-
-            if (mounted && _document != null && _controller != null) {
-              widget.params.onViewerReady?.call(_document!, _controller!);
-            }
           }
           _initialized = true;
         });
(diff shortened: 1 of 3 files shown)


espresso3389:
I have completely no idea but how much does PNG decompression impacts the actual rendering time do you think? If it is feasible enough, I want to implement some code on PdfImage side; like compact or such.

espresso3389:
Ah, I see what you're talking about. If we have PNG compressed version of page images, we can take 10x more cached pages even with the same memory cost. So we don't have to access the actual PDF any longer.

espresso3389:
If we do such compression, I want to do that on native code side. The structure will get simpler.

espresso3389:
> hmm.. What do you mean by the 'native code side'? If it is the place of that getBytes line, yes I know that is ulgy and costs extra unnecessary cpu cycle, but just for the sake of the callbacks and limit the scope of this PR.

I think passing png bytes instead of ui.Image is bad dicision even though it may perform well on certain situation. I want to encapsulate such things. And, PdfImage is already such a encapsulation on raw C++ memory buffered RGBA image. PdfImage.createImage does C++ memory RGBA image to ui.Image conversion.

So, we can do everything inside PdfImage and there might be chance to implement everything in C++ code or anyway inside PdfImage.

The user don't have to know the implementation detail. The underlying data can be anything; RGBA or PNG or JPEG.

espresso3389:
Each method call invokes pre-launched per-document Isolate using [BackgroundWorker.compute](https://github.com/espresso3389/pdfrx/blob/master/lib/src/pdfium/worker.dart#L37) so the cost is minimum.

espresso3389:
> That Isolate doesn't get launched until await _worker is called, meaning a new Isolate for every render.

No, it's not correct. For the first occurrence of (await _worker) may wait for the initialization but later calls just get the awaited-result immediately. No extra cost.
Only the cost there is asynchronously send the request to the worker thread.

espresso3389:
Anyway, the latest master commit b126686d611e6d1eaf7ccf840df7417ab25b4cf5 reduces total number of isolates used for first initialization of PdfDocument instances.

Now, only one isolate is created on PdfDocument initialization and it is passed to the PdfDocument and PdfDocument reuses the same isolate for every API calls including render method.

PR #136, opened 2024-04-22
guard for disposed document to avoid dead lock when a disposed worker is used

yhyh0:
probably not the best solution, please feel free to update it.


espresso3389:
1.0.57 contains the PR.

PR #137, opened 2024-04-22
To cancel unnecessary partial render

yhou0:


espresso3389:
One question, why do we have currentPartialCancelToken? Any solid reason to export this? I think it's too difficult to use such cancellation on real production apps.

espresso3389:
Anyway, 1.0.57 contains the PR.

PR #157, opened 2024-05-11
issue with flutter master channel

### LINKED ISSUE #156: issue with flutter master channel
I have tried the package with latest flutter master and failed to compile for function `dispatchSelectionEvent` on `lib/src/widgets/pdf_page_text_overlay.dart`

I found that they have introduced `SelectionEventType.selectParagraph` .. is to possible to add `default` case for `switch(event.type)` .. it will solve the issue for current master

I know its still not stable release but just want to highlight an issue that could be avoided once its pushed to beta and stable

raldhafiri:


raldhafiri:
fix #156

PR #176, opened 2024-06-18
interactionEndFrictionCoefficient param added

DevStaKat:
Hi, 

we added the param interactionEndFrictionCoefficient to viewer_params, so it can be set if wanted. Because the default value is not always the best.


espresso3389:
Your PR is anyway imcomplete and not compiable.

Anyway, I understood your intention. I'm now working on it.

PR #197, opened 2024-07-03
Libpdfium build 2 files found

HeropolisDa2ny:


espresso3389:
The changes are not acceptable.

android/build.gradle
espresso3389, line 1: `exclude` is too strong to incorporate into pdfrx's build.gradle. It's OK if you're the app owner but other apps have thier own configs. `pickFirst` is safer because it just accept one of libpdfium.so files but `exclude` may break all the dependencies in the worst case.

Issue #8, opened 2023-12-20
Execution failed for task ':pdfrx:mergeReleaseNativeLibs'. > A failure occurred while executing com.android.build.gradle.internal.tasks.MergeNativeLibsTask$MergeNativeLibsTaskWorkAction


felix-synorex:
I have trouble build the apk with pdfrx install.

Below is error I got
```
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':pdfrx:mergeReleaseNativeLibs'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.MergeNativeLibsTask$MergeNativeLibsTaskWorkAction
   > 2 files found with path 'lib/arm64-v8a/libpdfium.so' from inputs:
      - /opt/atlassian/pipelines/agent/build/build/pdfrx/intermediates/merged_jni_libs/release/out/arm64-v8a/libpdfium.so
      - /opt/atlassian/pipelines/agent/build/build/pdfrx/intermediates/cxx/RelWithDebInfo/1o3w592y/obj/arm64-v8a/libpdfium.so
     If you are using jniLibs and CMake IMPORTED targets, see
     https://developer.android.com/r/tools/jniLibs-vs-imported-targets
```

I have try to add packagingOptions like so in app level build.gradle but not working also.
```
android {
    
    packagingOptions {
        pickFirst 'lib/x86/libpdfium.so'
        pickFirst 'lib/x86_64/libpdfium.so'
        pickFirst 'lib/armeabi-v7a/libpdfium.so'
        pickFirst 'lib/arm64-v8a/libpdfium.so'
    }
//rest of flavors, compile, etc...
}
```

espresso3389:
I've tested `flutter build apk --release` on Windows and macOS with fresh flutter+Android Studio environment but no issues found during the build.

Could you please give me your `flutter doctor -v` result here?

felix-synorex:
I am runnning this from a `ghcr.io/cirruslabs/flutter:latest` docker image.

Here is the `flutter doctor -v` result:
```
[!] Flutter (Channel [user-branch], 3.16.4, on Ubuntu 22.04.2 LTS 5.10.102.1-microsoft-standard-WSL2, locale en_US.UTF-8)
    ! Flutter version 3.16.4 on channel [user-branch] at /sdks/flutter
      Currently on an unknown channel. Run `flutter channel` to switch to an official channel.
[...]

PR #204, opened 2024-07-04
Libpdfium build 2 files found (pickFirst)

HeropolisDa2ny:
Temp fix for the building bug of 2 files found for libpdfium.so

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -59,4 +59,11 @@ android {
     defaultConfig {
         minSdkVersion 21
     }
+
+    packagingOptions {
+        pickFirst 'lib/x86/libpdfium.so'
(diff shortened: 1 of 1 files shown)


sommen01:
It works for me

HeropolisDa2ny:
@espresso3389 are you fine to approve this PR or you would prefer to look for another solution ?

espresso3389:
As I descibed on https://github.com/espresso3389/pdfrx/issues/8#issuecomment-2205139356, this change is just a workaround and sometimes it gets very harmful. We need more discussion on it on #8 before accepting such PRs.

espresso3389:
After several hours of investigation, I finally decided to merge this anyway. There might be several side effects but the possibilities are so small. The risk can be acceptable.

PR #220, opened 2024-08-09
improve memory usage

jezell:
Have been seeing some crashes in Flutter web and noticed a lot of extra memory being used by pdfrx than needed. This pull request includes the following changes to help keep memory usage down:

- fix many potential memory leaks when widget is unmounted during rendering
- synchronizes rendering using document reference instead of widget state, so that multiple widgets referencing the same document reference will not run canvaskit renderer out of memory if they are quickly built and disposed before async rendering completes (current behavior can result in out of memory errors in web). 
- Better protection against dispose of images during async creation of image
- Properly enforce max / min scale so that final resolution of image is properly limited by min and max scale.
- Fixes exception thrown by interactive viewer when _alternativeFitScale is less than maxScale.

### DIFF
--- a/lib/src/pdf_api.dart
+++ b/lib/src/pdf_api.dart
@@ -399,15 +399,30 @@ abstract class PdfImage {
   /// Raw pixel data. The actual format is platform dependent.
   Uint8List get pixels;
 
+  bool _disposed = false;
+
   /// Dispose the image.
-  void dispose();
+  @mustCallSuper
+  void dispose() {
+    _disposed = true;
+  }
 
   /// Create [ui.Image] from the rendered image.
-  Future<ui.Image> createImage() {
+  Future<ui.Image> createImage() async {
+    if (_disposed) {
+      throw PdfException("Image is already disposed");
+    }
     final comp = Completer<ui.Image>();
+
     ui.decodeImageFromPixels(
         pixels, width, height, format, (image) => comp.complete(image));
-    return comp.future;
+
+    final result = await comp.future;
+    if (_disposed) {
+      result.dispose();
+      throw PdfException("Image was disposed during rendering");
+    }
+    return result;
   }
 }
 
--- a/lib/src/pdfium/pdfrx_pdfium.dart
+++ b/lib/src/pdfium/pdfrx_pdfium.dart
@@ -872,6 +872,7 @@ class PdfImagePdfium extends PdfImage {
 
   @override
   void dispose() {
+    super.dispose();
     malloc.free(_buffer);
   }
 }
--- a/lib/src/web/pdfrx_web.dart
+++ b/lib/src/web/pdfrx_web.dart
@@ -488,8 +488,6 @@ class PdfImageWeb extends PdfImage {
   final Uint8List pixels;
   @override
   PixelFormat get format => PixelFormat.rgba8888;
-  @override
-  void dispose() {}
 }
 
 class PdfPageTextFragmentWeb implements PdfPageTextFragment {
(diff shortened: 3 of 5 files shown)


espresso3389:
Your PR contains many changes. Please separate the changes into small PRs to discuss the things separately.

jezell:
@espresso3389 how would you like it to be split up? This is a set of changes which solve one problem, which is PDFRX causing our web app to crash by both leaking memory and using too much memory.

espresso3389:
At least your changes contains code changes not for memory leak. Some of them are for timing issues (mounted) or zoom calculation or... Even if they are really correct fixes, I could not accept them in a single PR.

espresso3389:
Your "memory leak fixes" code is anyway unacceptable because it breaks the compatibility with older codes. We should not introduce any new dispose codes on PdfImage. They should be on PdfViewer side or somewhere.

jezell:
The change to the dispose logic is because the image is never disposed at all ever and leaks if the widget is removed from the tree before the async work to generate the preview completes. The existing dispose logic clears the cached images when the widget is disposed, but if the work completes after widget.dispose is called, then the item is still placed in the cache even after the widget has unmounted and it will never be disposed and the pdf viewer leaks. 

This problem is made worse because things like the image scaling parameters are not handled properly, resulting in much larger images than needed being cached, as well as the poor coordination of the async work resulting in the problem happening in parallel across many instances of the viewer if you have more than one pdf viewer on the page. Imagine for example a ListView that contains PDF viewers, if you scroll around quickly in the ListView, sometimes it only takes 30 seconds to generate 2 GB worth of cached PDF images with pdfrx viewers and run the browser out of memory and crash flutter due to these bugs and leaks.

I do agree that the logic caching logic belongs somewhere else, maybe in a specialized PdfImageCache class or something rather than the PdfViewer. The code is definitely attempting to protect against some of these kinds of race conditions, but with the cache logic inside the viewer itself, it will be impossible to totally solve them. 

In any case, if you think fixing these problems is "unacceptable" it might be better for us to continue to work on a fork instead of attempting to fix them, as this was making our app crash every few minutes.

espresso3389:
Anyway the PR is too large. My first concern is that it does contain much things. We should do the things step by step.

And, I don't understand why the change need dispose method change on PdfImage. Is it critical to your scheme or strategy? I don't want that change at least.

The actual issue you want to solve seems reasonable, but the approach must be changed I think.

jezell:
The problem that that change is protecting against is real. This is how the leak happens:

A PdfImage is created by the pdf viewer.
PdfImage.createImage is called, but the call is async.
The pdf viewer is removed from the screen because it scrolls of the screen
The pdf viewer is disposed because it is removed from the screen
Disposing the pdf viewer calls PdfImage.dispose, but the createImage call to decodeImageFromPixels has not completed yet.
The decodeImageFromPixels call returns, creating a ui.Image, but this ui.Image will leak and never be disposed because the PdfImage has already been disposed and removed from the pdf viewer cache and will not be used. It throws an exception from the method because the work was cancelled before the image was returned, to make sure that the image doesn't end up getting used.

That change to the dispose logic makes sure that that the image doesn't leak in this case. 

Maybe you have a better way to improve the caching logic and fix the leaks than this PR, but for us this was the minimum set of changes to make our app stop crashing. It is a large set of related changes, but we found a large number of problems with the way memory is used and images are cached and all of these changes were made only to fix this problem. I do think that moving the cache itself somewhere outside the viewer would be a much better change, but I was trying to keep the changes as small as possible to get our app to stop crashing.

I have also opened up a pull request to the flutter engine itself to make memory leaks less likely to crash flutter on the web, but that will not be available until at least the next major flutter release.

espresso3389:
OK, I got your point. So, what we need is just another version of createImage. Adding a new method does not break the older apps. Please give me a little time to think more about the design to deal with async disposing procedure.

PR #262, opened 2024-11-28
Remove redundant check that breaks building on some systems

tymscar:
While trying to build any project that depends on pdfrx, I always get the following error:
```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
libc++
   linked by target "pdfrx" in directory /Users/tymscar/.pub-cache/hosted/pub.dev/pdfrx-1.0.91/src
```

Now after hours of trying everything, I stumbled upon this article from [developer.android.com](https://developer.android.com/ndk/guides/cpp-support#cmake) stating that:
`The default for CMake is c++_static.`

Because of this, I decided to remove ` ${libc++}` from `target_link_libraries` from pdfrx's `CMakeLists.txt` and it built without an issue after that.

I don't fully know why `find_library(libc++ libc++_static.a)` fails for me, and not for everybody, but considering it's redundant and it works without it for everyone, including me, I suggest this PR that removes the need.

### DIFF
--- a/android/CMakeLists.txt
+++ b/android/CMakeLists.txt
@@ -67,8 +67,6 @@ endif()
 file(REMOVE ${PDFIUM_LATEST_DIR})
 file(CREATE_LINK ${PDFIUM_LIBS_DIR} ${PDFIUM_LATEST_DIR} SYMBOLIC)
 
-find_library(libc++ libc++_static.a)
-
 set(pdfrx_bundled_libraries
   # Defined in ../src/CMakeLists.txt.
   # This can be changed to accommodate different builds.
@@ -78,4 +76,4 @@ set(pdfrx_bundled_libraries
 )
 
 target_include_directories(pdfrx PRIVATE ${PDFIUM_LATEST_DIR}/include)
-target_link_libraries(pdfrx PRIVATE ${PDFIUM_LASTEST_LIB_FILENAME} ${libc++})
+target_link_libraries(pdfrx PRIVATE ${PDFIUM_LASTEST_LIB_FILENAME})


espresso3389:
I'm very curious about your environment config including version of CMake, SDK, and NDK version because I'm not facing the issue.

And, I'm not sure whether your config is really safe and correct for everyone; only you and the other one are just complaining that.

tymscar:
cmake: `version 3.30.5`
SDK platform: Android 14
SDK built tools: 35.0.0 (but I've tried 6 different versions)
NDK: 26.1.10909125 (but I've also tried 25.2.8518653, 25.1.8937393, 23.1.7779620)

It's worth mentioning that my colleagues do not get this issue, and they are on the same versions as me, so I did not expect you to have it either, but that being said, the check is redundant, because it's setting libc++ to the static library, which is the default in CMake anyways. So you should not be breaking anything if you do that, and it makes the code shorter.

Have you tried running it with my suggestions on your machine?

espresso3389:
@sizeak Sorry, I didn't understand your message. Who do you mean by "you"?

So, should I merge it? Or any notable things?? I'm busy for my work now (it's almost december!!) and could not fully check the build behavior...

sizeak:
Hi @espresso3389!

Sorry, to clarify, I am a colleague of @tymscar. The "You" in "You only really need to consider finding..." is referring to an individual, i.e. "one", not a specific person. So "A library author only really needs to consider finding the path to the standard library when putting together a custom..." would be clearer wording.

The holidays are fast approaching aren't they! I'm sorry you're finding it busy, aren't we all!

sizeak:
As for whether you should merge it, I think that it is a safe and sensible change. My understanding of the Google documentation is that the call `find_library` call to locate the STL is unnecessary and we have tested this on multiple machines including Macs and Linux machines which could build with & without the change. 

The worst case is that you can roll it back and push a bug fix release, but that seems unlikely to be necessary.

tymscar:
Thank you so much

espresso3389:
I confirmed `flutter build appbundle` ran successfully on the following "fresh" env:

```
[✓] Flutter (Channel stable, 3.24.5, on Ubuntu 24.04.1 LTS 5.15.167.4-microsoft-standard-WSL2, locale C.UTF-8)
    • Flutter version 3.24.5 on channel stable at /home/kawasaki/work/flutter
    • Upstream repository https://github.com/flutter/flutter
    • Framework revision dec2ee5c1f (2 weeks ago), 2024-11-13 11:13:06 -0800
    • Engine revision a18df97ca5
    • Dart version 3.5.4
    • DevTools version 2.37.3

[✓] Android toolchain - develop for Android devices (Android SDK version 29.0.3)
    • Android SDK at /usr/lib/android-sdk
    • Platform android-29, build-tools 29.0.3
    • ANDROID_HOME = /usr/lib/android-sdk
    • Java binary at: /usr/bin/java
    • Java version OpenJDK Runtime Environment (build 21.0.5+11-Ubuntu-1ubuntu124.04)
    • All Android licenses accepted.

[✗] Chrome - develop for the web (Cannot find Chrome executable at google-chrome)
    ! Cannot find Chrome. Try setting CHROME_EXECUTABLE to a Chrome executable.

[✓] Linux toolchain - develop for Linux desktop
    • Ubuntu clang version 18.1.3 (1ubuntu1)
    • cmake version 3.28.3
    • ninja version 1.11.1
    • pkg-config version 1.8.1

[!] Android Studio (not installed)
    • Android Studio not found; download from https://developer.android.com/studio/index.html
      (or visit https://flutter.dev/to/linux-android-setup for detailed instructions).

[✓] Connected device (1 available)
    • Linux (desktop) • linux • linux-x64 • Ubuntu 24.04.1 LTS 5.15.167.4-microsoft-standard-WSL2

[✓] Network resources
    • All expected network resources are available.
```

espresso3389:
[1.0.92](https://pub.dev/packages/pdfrx/versions/1.0.92) is just released!

PR #264, opened 2024-12-03
Check for non-existent zoom element in PdfDest.params in some PDFs

remarkked:
Bookmarks in some PDFs (example attached) don't contain z (zoom) value, so the PdfDest.params.length = 2 ([x, y]), but _PdfViewerState._calcMatrixForDest tries to access PdfDest.params[2], so the error occurs when trying PdfViewerController.goToDest(PdfDest).
image
Requesting small fix for that

[pdf-example-bookmarks.pdf](https://github.com/user-attachments/files/17990854/pdf-example-bookmarks.pdf)


espresso3389 (approved):

espresso3389:
[1.0.93](https://pub.dev/packages/pdfrx/versions/1.0.93) contains the fix.

PR #293, opened 2025-01-11
Optimize PDF Text Search by Implementing Text Caching for Pages

abbashosseinii:
Description:
This PR introduces text caching for PDF pages to significantly improve search performance.

What changed?
Added a caching mechanism (_cachedText) to store text content of PDF pages after the first load. Subsequent searches for the same page now retrieve text from the cache instead of reloading it.

Why is this important?
This optimization drastically reduces search times for repeated queries. For example:

Initial search for a word: 29,288 ms

Subsequent search for the same word: 226 ms

Impact:
Improves user experience by making searches faster and more efficient, especially for repeated queries on the same page.


espresso3389:
OK, the reason for the change is acceptable.

PR #349, opened 2025-04-27
Fix crash on window resize on desktop (Image.dispose error)

abbashosseinii:
## Problem
When resizing the window on desktop, the following error occurred:
[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: 'dart:ui/painting.dart': Failed assertion: line 1979 pos 12: '<optimized out>': is not true.

This happened because the old image was being disposed after a new image creation without handling potential failures.

## Solution
Wrapped the image creation and disposal logic in a try-catch block to ensure safe error handling.  
If an exception occurs during image creation, it is now logged, and resources are still properly disposed without causing crashes.

## Notes
- Added a log message for easier debugging (`developer.log`).
- This fix improves the stability of resizing behavior on desktop.

### DIFF
--- a/lib/src/widgets/pdf_widgets.dart
+++ b/lib/src/widgets/pdf_widgets.dart
@@ -211,6 +211,7 @@ class _PdfPageViewState extends State<PdfPageView> {
 
   @override
   void dispose() {
+    _image?.dispose();
     _cancellationToken?.cancel();
     super.dispose();
   }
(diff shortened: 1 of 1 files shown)


espresso3389 (approved):

espresso3389:
pdfrx [1.1.24](https://pub.dev/packages/pdfrx/versions/1.1.24) contains the fix.

PR #364, opened 2025-05-15
fix: blank pdf on Windows when restore window from minimize

letrungdo:
on Windows when minimize window to taskbar, after click to app icon in taskbar to restore window
=> pdf will not display anymore


espresso3389:
I didn't reproduce your issue but the changes seems OK.

PR #410, opened 2025-07-08
Fix build issue on pdf_page_text_overlay.dart

hzaun:
Fix for breaking change on flutter  3.32.0+, where 
SelectionHandler interface has added two new abstract members:

-> int get contentLength;

-> SelectedContentRange? getSelection();


espresso3389:
You can see these members were already implemented [here](https://github.com/espresso3389/pdfrx/blob/4a03cc84e4d29466974ae2df2c5ad047cfad1277/lib/src/widgets/pdf_page_text_overlay.dart#L551).

See 2b20e2aa288e076c3a6e906f2d0dbf52cec033c0 (pdfrx v1.1.1)

PR #481, opened 2025-09-25
preserve precise position on page for layout and view size changes

enhancient:
In the current implementation of PdfView, when layout or view size changes, _goToPage() is called which will likely change the zoom and position of the page that is currently being viewed. If the user was in the midst of reading the document, they must re-orient and find their previous position, which is undesirable.

This PR includes a change that retains the user's precise position (relative to the top left of the page) when there are size changes (for example rotation of a device), and tries to retain the current zoom level where possible. It will also keep the document from going out of bounds. Even if the document layout changes (for example different layouts might be setup for landscape vs portrait viewing), the position is maintained according to the previous position in the PDF document coordinate space. The attached videos show a before and after this change.

Some implementation details:

- Any current animations in the InteractiveViewer are stopped so that these animations do not interfere with any changes to the matrix we make
- a new _goToPosition method has been added which goes to a specific offset within the document
- Partial images in the cache are removed as these may appear in the incorrect position
- when the layout or size changes, _updateLayout  finds the closest point within the PdfPage to the top left corner of the screen in the old page layout and translates the position of this point in Pdf document space to the new page layout.
- the onViewSizeChanged method in the viewer example has been removed, as this is redundant with this change

https://github.com/user-attachments/assets/5e9157af-91e7-4d63-92a5-8c61f25bef0e
https://github.com/user-attachments/assets/f09bdc62-de1a-4fd9-a6d6-ad5421006e23

### DIFF
--- a/packages/pdfrx/lib/src/widgets/interactive_viewer.dart
+++ b/packages/pdfrx/lib/src/widgets/interactive_viewer.dart
@@ -60,6 +60,55 @@ typedef InteractiveViewerWidgetBuilder = Widget Function(BuildContext context, Q
 /// {@end-tool}
 @immutable
 class InteractiveViewer extends StatefulWidget {
+
+  /// Create InteractiveViewer with animation control capability
+  InteractiveViewer.withAnimationControl({
+    required Widget child,
+    Key? key,
+    Clip clipBehavior = Clip.hardEdge,
+    PanAxis panAxis = PanAxis.free,
+    EdgeInsets boundaryMargin = EdgeInsets.zero,
+    bool constrained = true,
+    double maxScale = 8.0,
+    double minScale = 0.8,
+    double interactionEndFrictionCoefficient = _kDrag,
+    GestureScaleEndCallback? onInteractionEnd,
+    GestureScaleStartCallback? onInteractionStart,
+    GestureScaleUpdateCallback? onInteractionUpdate,
+    bool panEnabled = true,
+    bool scaleEnabled = true,
+    double scaleFactor = kDefaultMouseScrollToScaleFactor,
+    TransformationController? transformationController,
+    Alignment? alignment,
+    bool trackpadScrollCausesScale = false,
+    void Function(PointerScrollEvent event)? onWheelDelta,
+    ScrollPhysics? scrollPhysics,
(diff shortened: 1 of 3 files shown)


espresso3389:
@enhancient BTW, which dev environment do you normally working on?
Everytime I receive your PR, the formatting is broken and  pub.dev complains about the difference (and it's the reason why I introduced pana for checking code integrity... and it's also broken so far).

enhancient:
> @enhancient BTW, which dev environment do you normally working on? Everytime I receive your PR, the formatting is broken and pub.dev complains about the difference (and it's the reason why I introduced pana for checking code integrity... and it's also broken so far).

vscode, but I've recently been using Claude Code, and it looks like it wasn't running dart format,  but I've now added a pre-commit hook so hopefully going forward it shouldn't be an issue. I've pushed some formatting fixes and it's passed now.

packages/pdfrx/lib/src/widgets/interactive_viewer.dart
```diff
@@ -398,6 +447,13 @@ class InteractiveViewer extends StatefulWidget {
   // This value was eyeballed to give a feel similar to Google Photos.
   static const double _kDrag = 0.0000135;
 
+  /// GlobalKey to access the InteractiveViewer state for animation control
+  static final GlobalKey<_InteractiveViewerState> _globalKey = GlobalKey<_InteractiveViewerState>();
```
espresso3389, line 451: It's too bad to use static GlobalKey on InteractiveViewer. It makes InteractiveViewer almost singleton.

espresso3389:
I've once merged the changes because of many changes I want to apply on the changes. (Without doing so, the PRs will be unable to merge like the previous large PR by you)

espresso3389:
The update for this issue has been released in [v2.1.20](https://pub.dev/packages/pdfrx/versions/2.1.20).

- Added `PdfViewerParams.scrollPhysics` and `PdfViewerParams.scrollPhysicsScale` so custom `ScrollPhysics` can control panning and pinch-zoom interactions.

Release notes: https://github.com/espresso3389/pdfrx/blob/master/packages/pdfrx/CHANGELOG.md#2120

Written by Codex Agent

PR #485, opened 2025-09-27
Add scrollPhysicsScale and FixedOverScrollPhysics

enhancient:
I forgot to add the scrollPhysicsScale argument, which enables different scrollPhysics for scaling and an associated FixedOverScrollPhysics which enables ClampingScrollPhysics for panning and a modified BouncingScrollPhysics for overshoot and snapback to replicate the behavior seen on several popular PDF readers on Android.  I think we should add this change now before releasing the scrollPhysics feature.

There is also a helper function to select the appropriate scrollPhysics for the platform: PdfViewerParams.getScrollPhysics(context). Currently it chooses the FixedOverScrollPhysics() for Android, and then falls through to  ScrollConfiguration.of(context).getScrollPhysics(context). A decision needs to be made as to whether FixedOverScrollPhysics() should apply for other platforms such as Windows.

I also made some changes to the documentation to reflect these changes.

### DIFF
--- a/packages/pdfrx/lib/src/widgets/pdf_viewer_params.dart
+++ b/packages/pdfrx/lib/src/widgets/pdf_viewer_params.dart
@@ -1,3 +1,4 @@
+import 'dart:io';
 import 'dart:ui' as ui;
 
 import 'package:flutter/foundation.dart';
@@ -71,6 +72,7 @@ class PdfViewerParams {
     this.behaviorControlParams = const PdfViewerBehaviorControlParams(),
     this.forceReload = false,
     this.scrollPhysics,
+    this.scrollPhysicsScale,
   });
 
   /// Margin around the page.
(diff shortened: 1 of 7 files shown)


packages/pdfrx/lib/src/widgets/pdf_viewer_params.dart
```diff
@@ -539,13 +541,35 @@ class PdfViewerParams {
 
   /// Scroll physics for the viewer.
   ///
-  /// If null, it works like [ClampingScrollPhysics] on all platforms.
```
enhancient, line 542: @espresso3389 technically the internal InteractiveViewer physics is quite different than even ClampingScrollPhysics when panning, so I've updated the documentation on this. InteractiveViewer physics was actually designed more for panning images rather than scrolling through a document like a PDF, so the friction simulation settles much quicker.

packages/pdfrx/lib/src/widgets/pdf_viewer_params.dart
```diff
@@ -539,13 +541,35 @@ class PdfViewerParams {
...
+  /// does not allow zooming beyond the min/max scale, and flings on panning come to rest quickly relative to
+  /// Scrollables in Flutter (such as SingleChildScrollView).
+  ///
+  /// A convenience function [getScrollPhysics] is provided to get platform-specific default scroll physics.
+  /// If you want no overscroll, but still want the physics for panning to be similar to other Scrollables,
+  /// you can use
+  /// [ClampingScrollPhysics()].
   ///
   /// If the value is set non-null, it disables [normalizeMatrix].
-  /// If you set [boundaryMargin] to `EdgeInsets.all(double.infinity)`, it effectively disables [scrollPhysics].
```
enhancient, line 546: @espresso3389 It's true that setting boundaryMargin to infinity will disable boundary clamping, but the physics for panning will still remain effective and according to the scrollPhysics set.

espresso3389: I'll update the comment.

enhancient:
@espresso3389 also, packages/pdfrx/lib/src/widgets/pdf_viewer_scroll_thumb.dart should be updated to consider boundaryMargin values that are non-zero/infinite. Did you want to do that?

espresso3389:
> @espresso3389 also, packages/pdfrx/lib/src/widgets/pdf_viewer_scroll_thumb.dart should be updated to consider boundaryMargin values that are non-zero/infinite. Did you want to do that?

~~If you can implement it, please do that.~~

Ah, if we want to do that, we should introduce some mechanim to pass params to thumbs??

packages/pdfrx/lib/src/widgets/pdf_viewer_params.dart
enhancient: I already updated this in the commit 😄

enhancient:
@espresso3389 looks like pana is failing with:

> Generating docs for library pdfrx.dart from package:pdfrx/pdfrx.dart...
> ERR:
> dartdoc 8.3.4 (/tmp/pana.1758943442705.DHNKMT/pub-cache/global_packages/dartdoc/bin/dartdoc.dart-3.9.2.snapshot) failed: Stack Overflow
> 

Do you know why this might be?

enhancient:
> Ah, if we want to do that, we should introduce some mechanim to pass params to thumbs??

It's accessible through widget.controller.params.boundaryMargin because the controller is supplied as an argument - do you think this is an acceptable way to access params?

espresso3389:
> It's accessible through widget.controller.params.boundaryMargin because the controller is supplied as an argument - do you think this is an acceptable way to access params?

Yes, of cource.

espresso3389:
The update for this issue has been released in [v2.1.20](https://pub.dev/packages/pdfrx/versions/2.1.20).

- Added `PdfViewerParams.scrollPhysics` and `PdfViewerParams.scrollPhysicsScale` so custom `ScrollPhysics` can control panning and pinch-zoom interactions.

Release notes: https://github.com/espresso3389/pdfrx/blob/master/packages/pdfrx/CHANGELOG.md#2120

Written by Codex Agent

PR #502, opened 2025-10-16
InteractiveViewer ScrollPhysics pinch-zoom centering fixes

enhancient:
When scrollPhysics was specified, pinch-zooming up from minScale but when the document was still smaller than the viewport was not always centering the document properly at the end of the gesture. This fix addresses that.


espresso3389:
The fix for this issue has been released in [v2.2.2](https://pub.dev/packages/pdfrx/versions/2.2.2).

This release addresses the InteractiveViewer ScrollPhysics pinch-zoom centering issues where zooming up from minScale with a document smaller than the viewport was not properly centering the document at the end of the gesture.

🤖 Written by Claude Code

Issue #347, opened 2025-04-24
Current page not correctly detected in PDFs with landscape orientation


andersonmendesgd:
First of all, I'd like to thank and congratulate you for the amazing work on pdfrx. This component has been incredibly useful in our project, and the attention to detail and performance is truly appreciated. 🙌

When using PdfViewer to view PDFs with pages in landscape orientation (e.g., PowerPoint presentations), the onPageChanged callback does not accurately reflect the currently visible page. This leads to inconsistent behavior, such as incorrect page numbers and issues when relying on the current page value.

**Code snippet used:**
```dart
PdfViewer.data(
  snapshot.data!,
  sourceName: widget.url,
  controller: pdfController,
  params: PdfViewerParams(
    backgroundColor: Colors.grey[200]!,
    enableTextSelection: true,
    onPageChanged: (pageNumber) => onPageChanged(pageNumber!),
    pageOverlaysBuilder: (context, pageRect, page) {
      return [
        Align(
          alignment: Alignment.bottomCenter,
          child: Text(page.pageNumber.toString(), style: const TextStyle(color: Colors.red)),
        ),
      ];
    },
    linkHandlerParams: PdfLinkHandlerParams(
      onLinkTap: (link) {
        if (link.url != null) {
          openLink(link.url!);
        } else if (link.dest != null) {
          pdfController.goToDest(link.dest);
        }
      },
      linkColor: Colors.transparent,
    ),
    viewerOverlayBuilder: (context, size, handleLinkTap) => [
      GestureDetector(
        behavior: HitTestBehavior.translucent,
        onTapUp: (details) {
          handleLinkTap(details.localPosition);
        },
        onDoubleTap: () {
          pdfController.zoomUp(loop: true);
        },
        child: IgnorePointer(child: SizedBox(width: size.width, height: size.height)),
      ),
      PdfViewerScrollThumb(
        controller: pdfController,
        orientation: ScrollbarOrientation.right,
        thumbSize: const Size(40, 25),
        thumbBuilder: (context, thumbSize, pageNumber, controller) => Container(
          color: Colors.black,
          child: Center(
            child: Text(
              pageNumber.toString(),
              style: const TextStyle(color: Colors.white),
            ),
          ),
        ),
      ),
      PdfViewerScrollThumb(
        controller: pdfController,
        orientation: ScrollbarOrientation.bottom,
        thumbSize: const Size(80, 10),
[...]

PR #507, opened 2025-10-21
_guessCurrentPageNumber() enhancement

enhancient:
Enhanced  current page number detection algorithm to provide more accurate and stable page tracking across different document layouts and viewing scenarios. Should address issues such as #347. Also caters for scenarios where minZoom is set < 1 and multiple pages might be visible at once. 

At a high level it follows PDF.js sticky page approach (https://github.com/mozilla/pdf.js/blob/master/web/pdf_viewer.js#L1757) for the normal case with added edge page detection, and when dealing with complex layouts or zoomed out to reveal many pages, it reverts to a scroll percentage mode to provide more useful/expected results.

Key Improvements
1. Primary Axis Visibility Calculation
New _calcPrimaryAxisVisibility() method that calculates page visibility along the primary scroll axis which is
more accurate than area-based intersection for determining which page the user is viewing
Works correctly for both horizontal and vertical layouts
2. Layout Type Detection
New _isSimpleLayout() method distinguishes between sequential and facing page layouts
Detects when pages overlap along the scroll axis (facing pages)
Enables different behaviors for simple vs. complex layouts
Note, this would be updated if/when #505 is implemented.
3. Scroll Percentage Mode
Calculates page number based on scroll position rather than visibility when appropriate
Automatically switches to scroll percentage mode when 3+ pages are fully visible or in complex layouts
5. Enhanced Sticky Mode
Keeps current page stable when fully visible
Edge detection: prefers first/last page when also fully visible
Proximity-based tie-breaking when multiple pages have equal visibility


psh-developers:
please upload this issue fix

Issue #430, opened 2025-07-26
[Web] Pdfium memory not cleared when pdf is disposed


defuncart:
As a developer, when I dispose a pdf, I would also expect that memory consumed by Pdfium would also be disposed.

## Problem

This issue is as mentioned in https://github.com/espresso3389/pdfrx/issues/184, however creating a separate issue for web as the implementation is different from native io platforms.

This can be easily reproduced by modifying the example project to initially display an empty screen with a button pushing the home page. When the home page is popped, the pdf itself is disposed, however Pdfium is still in memory.

https://github.com/user-attachments/assets/72875dda-6b24-4950-89be-7b2062eb065a

Once a pdf is displayed for the first time, roughly 40Mb memory is added to the heap, while a web worker is spun up. When the pdf is disposed, the memory from Pdfium is not released. On mobile devices running a web app, this is also around 30-40Mb.

## What I Tried

By modifying [`packages/pdfrx/assets/pdfium_client.js`](https://github.com/espresso3389/pdfrx/blob/master/packages/pdfrx/assets/pdfium_client.js), I was able to shut down the web worker as follows:

```js
globalThis.disposePdfium = function () {
  worker?.terminate();
  globalThis.worker = null;
  globalThis.PdfiumWasmCommunicator = null;
  globalThis.pdfiumWasmWorkerUrl = null;
};

globalThis.PdfiumWasmCommunicator = (function () {
  globalThis.worker = new Worker(globalThis.pdfiumWasmWorkerUrl);
...
```

```dart
@JS('disposePdfium')
external void disposePdfium();

@override
Future<void> dispose() async {
  ...
  disposePdfium();
}
```

however this did not free up much memory, and in fact broke usage on second viewing as the communicator etc. was null.

I also removed

```
<script type="module" charset="utf-8" async="" src="http://localhost:57301/assets/packages/pdfrx/assets/pdfium_client.js"></script>
```

which is injected into `head`, however the memory remained. I also tried to null all vars from dart's side, and experimented with the initialization flow. I guess Pdfium does something under the hood which would require much deeper analysis.

In short, I could not find out a way to shut down Pdfium and re-initialize from start on subsequent viewings.

## Alternatives

I tested [pdfx](https://github.com/ScerIO/packages.flutter/tree/main/packages/pdfx) which correctly disposes memory used by pdf.js, however there are performance issues when scrolling on web (https://github.com/ScerIO/packages.flutter/issues/314).

[...]

Issue #184, opened 2024-06-21
Memory not clear when pop from screen


FeofanGreek:
Hello. We noticed an unpleasant feature for iOS. After loading any PDF file using a controller in the widget, the memory is not cleared when leaving the screen. We collect more then 200 MBytes of garbage
I wanted to add a dispose controller or a cancel controller in dispose block of StateFullWidget, but the controller does not have such a methods..

espresso3389:
Could you please explain your issue in detail?

> After loading any PDF file using a controller in the widget

I cannot understand the word "controller" in the sentence. At least, PdfViewerController does not have mechanism to load PDF files.

FeofanGreek:
I mean PdfViewerController, but I may be wrong in describing the mechanism of its operation. I just assumed that it should be responsible for clearing memory after the user leaves the sc
![Снимок экрана 2024-06-22 в 07 57 06](https://github.com/espresso3389/pdfrx/assets/57085995/9d3c2a0a-8385-4f16-b341-cb89c7db6978)

![Снимок экрана 2024-06-22 в 07 56 46](https://github.com/espresso338
![Снимок экрана 2024-06-22 в 07 57 23](https://github.com/espresso3389/pdfrx/assets/57085995/f11b565f-2c7b-4edf-a6ef-1b682d6dd4e2)
9/pdfrx/assets/57085995/4e72f5d5-c0e5-4bf2-ba6b-da5de833018a)
reen with the PDF file downloaded and displayed. I have attached 3 screenshots from our application to the message, as an example of showing memory consumption.

espresso3389:
I tested pdfrx with the following code and it reveals that only the first time load caches "something" on memory but repeatedly open/close the PDF page does not consume more. I think the document itself seems correctly closed (Actually, it's opened and closed everytime use some Pdfium functions).

The problem is with some resource caches by Pdfium. The possible cached resources are fonts and color profiles though I don't know the actual implementation.

```dart
import 'package:flutter/material.dart';
import 'package:pdfrx/pdfrx.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  // This widget is the root of your application.
  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        colorScheme: ColorScheme.fromSeed(seedColor: Colors.deepPurple),
        useMaterial3: true,
      ),
[...]

PR #519, opened 2025-11-08
doc: Update README with performance considerations

defuncart:
Follow up to #430 and #184.


espresso3389:
README.md is already too much for the users; I'm thinking to reduce the contents on README.md actually. I don't want to write such low priority info., on README.md.

If we want to note about the issue, it should be placed under doc/.

And I think it's not the concern for the most of users anyway.

PR #523, opened 2025-11-10
Magnifier enhancements to support customization

enhancient:
This PR includes enhancements to the text selection magnifier to enable more fine grained control over positioning and visibility to support in/out animations. An example of a more complex magnifier in the form of an experimental iOS Liquid Glass magnifier appears below.

Key changes:

Non-breaking:

`PdfTextSelectionParams`
- `onSelectionHandlePanStart` - Callback fired when handle drag begins
- `onSelectionHandlePanUpdate` - Callback fired during handle drag with delta
- `onSelectionHandlePanEnd` - Callback fired when handle drag ends

`PdfViewerSelectionMagnifierParams` 
- `shouldShowMagnifier` - Optional callback for explicit magnifier visibility control
- `animationDuration` - Configurable position animation duration
- `calcPosition` - Optional callback for custom magnifier positioning

Breaking:

`PdfViewerGetMagnifierRectForAnchor`
- `pointerPosition` - added to enable positioning of magnifier content according to the pointer position. 
Using pointerPosition provides a smoother movement of the magnifying glass.

`PdfViewerMagnifierBuilder`
- `pointerPosition`  - as above 
- `magnifierPosition` - to enable animating the magnifier to the final position from another position

https://github.com/user-attachments/assets/f3e17fbc-3a39-4d68-b51d-f61c7f497c71

PR #528, opened 2025-11-11
Magnifier customization enhancements

enhancient:
This resubmit fixed the changed position of the magnifying glass of the previous PR #523, and improves some function names and documentation for consistency. 

Note I've added a commented out example of using the new parameters in main.dart but you likely do not want to include this in master - they are just there for testing. The example shows the following key enhancements that are possible with this PR:

- Stable movement of the magnifier by tracking against the pointer
- Stable movement of the content in the magnifying glass to match expected behaviour
- Clamping of content in the magnifying glass when it reaches viewport boundaries
- Ability to control the visibility of  the magnifying glass to enable in/out animations

### DIFF
--- a/packages/pdfrx/lib/src/widgets/pdf_viewer_params.dart
+++ b/packages/pdfrx/lib/src/widgets/pdf_viewer_params.dart
@@ -967,7 +967,9 @@ typedef PdfViewerSelectionHandlePanUpdateCallback = void Function(PdfTextSelecti
 /// Callback for when a selection handle pan ends
 typedef PdfViewerSelectionHandlePanEndCallback = void Function(PdfTextSelectionAnchor anchor);
 
-/// Text selection
+/// Interface for text selection information.
+///
+/// To perform text selection actions, use [PdfTextSelectionDelegate].
 abstract class PdfTextSelection {
   /// Whether the text selection is enabled by the configuration.
   ///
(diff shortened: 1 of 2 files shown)


packages/pdfrx/lib/src/widgets/pdf_viewer_params.dart
```diff
@@ -1018,6 +1045,9 @@ class PdfViewerSelectionMagnifierParams {
     this.builder,
     this.shouldBeShownForAnchor,
     this.maxImageBytesCachedOnMemory = defaultMaxImageBytesCachedOnMemory,
+    this.calcPosition,
+    this.shouldShowMagnifier,
```
espresso3389: The inconsistency between existing `shouldBeShownForAnchor` and `shouldShowMagnifier` naming rules. I'm thinking renaming of the existing `shouldBeShownForAnchor`. On the other hand, the problem is that they are very similar and confusing. We should do something on that (at least, explain the difference on document)...

espresso3389:
I'm going to merge this and PR #505 soon as a major upgrade because even this PR will have some API compatibility breakage. And I'm not sure which order should I merge the PRs.

espresso3389:
[pdfrx 2.2.11](https://pub.dev/packages/pdfrx/versions/2.2.11) contain the feature!

Issue #532, opened 2025-11-13
Maginifier content location is calculated incorrectly


espresso3389:
Related: #528
CC: @enhancient 

Apparently, there are mistakes in handling of global/local offsets.

https://github.com/user-attachments/assets/40e52ca5-ca5e-4514-9206-b72f485d810b

espresso3389:
Is it just a example issue??

enhancient:
Good catch - and yes, it's just an issue with the example in main.dart.  PR with fix sent.

Do you really want the magnifier example uncommented in main.dart though, because calcPosition for example isn't taking into consideration all text directions like the builtin calPosition is - maybe we should update the example, or even include these changes in the default calcPosition? I think more testing is probably required before that though.

espresso3389:
I found the cause of the issue.
`clampedPointerPosition` is viewer local, so 

https://github.com/espresso3389/pdfrx/blob/f6614c3c3f831f955e86a72f1333c034d4dfe1cc/packages/pdfrx/example/viewer/lib/main.dart#L327

espresso3389:
~~I could determine whether we should enable the code or not~~, but at least, the example code should work without any error.

I'll comment out the code on release anyway.

espresso3389:
As a side note, f12ffd3e8549965005f983ae9896d4fd5f6994d5 adds Add PdfViewerController.localToDocument/documentToLocal that can be used in such case.

espresso3389:
The fix for this issue has been released in [v2.2.11](https://pub.dev/packages/pdfrx/versions/2.2.11).

The magnifier content location calculation has been corrected by refactoring the coordinate conversion methods. Additionally, two new public methods have been added to `PdfViewerController` for coordinate conversion:
- `localToDocument(Offset local)` - Converts local widget coordinates to PDF document coordinates
- `documentToLocal(Offset document)` - Converts PDF document coordinates to local widget coordinates

These methods provide better control over coordinate transformations and can be useful for custom overlays or interactions with the PDF viewer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #533, opened 2025-11-13
Fix/532 magnifier position

enhancient:
Fix for #532


espresso3389:
Sorry, I did the same thing on master.

Issue #536, opened 2025-11-13
Feature Request: Add loadingBuilder and errorBuilder to PdfDocumentViewBuilder


psh-developers:
Currently, PdfDocumentViewBuilder does not provide an option to handle loading or error states while the PDF document is being loaded or if loading fails.

It would be very useful to have optional `loadingBuilder` and `errorBuilder` parameters.

That way, developers can show a custom loading indicator or an error message instead of leaving the view blank or handling it externally.

### Example usage:
```dart
PdfDocumentViewBuilder.asset(
  'assets/sample.pdf',
  loadingBuilder: (context) => const Center(
    child: CircularProgressIndicator(),
  ),
  errorBuilder: (context, error) => Center(
    child: Text('Failed to load PDF: $error'),
  ),
  builder: (context, document) => ListView.builder(
    itemCount: document?.pages.length ?? 0,
    itemBuilder: (context, index) {
      return Container(
        margin: const EdgeInsets.all(8),
        height: 240,
        child: Column(
          children: [
            SizedBox(
              height: 220,
              child: PdfPageView(
                document: document,
                pageNumber: index + 1,
              ),
            ),
            Text('${index + 1}'),
          ],
        ),
      );
    },
  ),
);
```

### Expected behavior:
- While the PDF is being loaded, the `loadingBuilder` widget should be displayed.
- If the PDF fails to load (e.g., invalid path, network error, corrupted file), the `errorBuilder` widget should be displayed.
- Once loaded successfully, the `builder` callback should render the document.

psh-developers:
hey @espresso3389 
Please do it today :D

espresso3389:
> hey @espresso3389
> Please do it today :D

I live in Tokyo and only 1 hour remaining today😪

psh-developers:
Hey @espresso3389, i really appreciate your hard working on this amazing lib
AI helped me to do it
and it worked perfectly
if you see it correct please tell me

```dart
class PdfDocumentViewBuilder extends StatefulWidget {
  const PdfDocumentViewBuilder({
    required this.documentRef,
    required this.builder,
    this.loadingBuilder,
    this.errorBuilder,
    super.key,
  });

  PdfDocumentViewBuilder.asset(
    String assetName, {
    required this.builder,
    this.loadingBuilder,
    this.errorBuilder,
    super.key,
    PdfPasswordProvider? passwordProvider,
    bool firstAttemptByEmptyPassword = true,
    bool useProgressiveLoading = false,
[...]

PR #537, opened 2025-11-13
Feature Request: Add loadingBuilder and errorBuilder to PdfDocumentVi…

espresso3389:
@psh-developers

Based on your issue #536 and code, I created a PR that contains the changes almost identical to your code.
The difference is that the code does not change the existing behavior if the user don't specify `loadingBuilder` and/or `errorBuilder`.


G705-Ghilan:
please update this on pub.dev as fast as possible
and thanks for your hard working @espresso3389

PR #564, opened 2025-12-10
Switch to package:hooks

mosuem:
Just an idea - you could also use `package:native_toolchain_c` to build it from scratch instead of downloading the dylib, if the user wishes to do so.

This should help with removing much of the complexity. I believe `pdfium_flutter` could be removed in the future.

### DIFF
--- a/packages/pdfium_dart/pubspec.yaml
+++ b/packages/pdfium_dart/pubspec.yaml
@@ -15,7 +15,6 @@ dependencies:
   ffi: ^2.1.4
   hooks: ^1.0.0
   http: ^1.5.0
-  native_toolchain_c: ^0.17.4
   path: ^1.9.1
 
 dev_dependencies:


espresso3389:
I've already evaluated hooks on Dart and I think it's still not mature. It does not resolve things I want to solve. At least, I could not accept it now.

mosuem:
That's very interesting to know - any specific issues you are facing?

This feature is out of experimental now, so should be stable enough and able to solve most use cases.

espresso3389:
Actually, in theory, your current, just downloading code works well (actually I didn't test it). But if so, it's almost identical to current situation and no need to take risk to upgrade the current implementation.

And, if we try to build pdfium on the local machine, it is not a easy task. It needs many toolchains, python, gclient, cmake, ninja... And also it takes so much time to build. Furthermore, pdfium's build script often breaks due to source changes; for me, it's too much painful to catch up the lastest source code.

Recently, I at least find out the way to use universal single xcframework for iOS/macOS, and that's the reason why we have `pdfium_flutter`. Without this, we would still depend on fragile my hand-crafted build script for iOS/macOS. And, hooks currently still lacks some vital support for iOS. Not usable at all for iOS.

--

BTW, for macOS, the current implementation still have issues on signing. It seem Dart could not run with pre-signed dylib downloaded from [bblanchon/pdfium-binaries](https://github.com/bblanchon/pdfium-binaries). I should some more work for it.

mosuem:
Thanks for the feedback!

The direct benefit would be no need for an async init step.

i don't know much about mac/ios builds, but I would expect this to work out of the box there as well. cc @dcharkes for insights.

I also like downloading precompiled binaries more, see for example https://github.com/unicode-org/icu4x/blob/main/ffi/dart/hook/build.dart for how the ICU4X package does it.

espresso3389:
I'm now re-thinking the things and it may be planned on [pre-3.0](https://github.com/espresso3389/pdfrx/discussions/534) versions of pdfrx_engine or such. At least, I don't want to introduce such that may affect compatibility with existing code base.

PR #580, opened 2025-12-30
Smooth Scroll & Zoom interactions (optional)

quaaantumdev:
Currently, mouse wheel scrolling and trackpad zooming in `pdfrx` feel "janky" because they translate inputs directly into instant matrix updates. While functional, it lacks the smooth, organic feel users expect from modern web browsers or native PDF readers. Rapid inputs (like a fast scroll wheel) result in jarring visual jumps rather than fluid motion.

We wanted to bring the high-quality, physics-based animation feel to `pdfrx` without forcing a breaking change on existing users.

### The Solution: Pluggable Interaction Architecture
To solve this, we refactored the input handling logic out of the core widget and into a delegate pattern:

*   **`PdfViewerScrollInteractionDelegate`**: An abstract base class that defines how `pan` and `zoom` requests are executed.
*   **`PdfViewerScrollInteractionDelegateProvider`**: A factory pattern used in `PdfViewerParams`. We use a provider class (instead of a simple callback) to ensure `const` equality works correctly, preventing unnecessary delegate recreation/disposals during widget rebuilds.

We implemented two concrete versions:
1.  **`...ProviderInstant` (Default)**: Replicates the exact behavior of the current version. Inputs result in immediate matrix updates. **This ensures no breaking changes for existing users.**
2.  **`...ProviderPhysics` (New)**: Uses `Ticker`s and exponential decay math to animate transitions. It handles "additive" animation, meaning if you scroll while an animation is running, the velocity accumulates smoothly rather than resetting. This feels significantly better on Desktop and Web.

### Additional Improvements
While refactoring the input logic, we added:
*   **Shift + Scroll Support**: Holding `Shift` while scrolling the mouse wheel now triggers horizontal scrolling, a standard desktop interaction pattern that was previously missing.
*   **Trackpad Pinch / Browser Zoom**: We improved how `PointerScaleEvent` (e.g., trackpad pinch or Ctrl+Scroll in some browsers) is handled, routing it through the same delegate system for consistent behavior.

### How to Use
To enable the new smooth scrolling, users simply pass the physics provider in the params.

```dart
PdfViewer.asset(
  'assets/document.pdf',
  params: PdfViewerParams(
    // Enable the physics-based interaction
    interactionDelegateProvider: const PdfViewerScrollInteractionDelegateProviderPhysics(),
    
    // Optional: Tune sensitivity (see below)
    scrollByMouseWheel: 1.0, 
    scaleByPointerScale: 0.3,
  ),
);
```

### Configuration & Recommended Values
We introduced a new parameter `scaleByPointerScale` to `PdfViewerParams`.
*   This acts as a sensitivity multiplier for trackpad pinch events and Ctrl+Scroll zooming.
*   **Default:** `1.0` (Preserves legacy behavior/sensitivity).

**Our Recommendation:**
The default sensitivity values felt to aggressive when zooming and to little when scrolling to us. We found these values provide a better feel:
*   `scrollByMouseWheel`: **1.0** (Default is 0.2)
*   `scaleByPointerScale`: **0.3** (Default is 1.0)

### Closing
Thanks for building `pdfrx`! It is a fantastic library. We hope this architectural improvement helps make it the go-to choice for high-fidelity Flutter PDF viewing on all platforms. We would love to see this merged!

### DIFF
--- a/packages/pdfrx/lib/pdfrx.dart
+++ b/packages/pdfrx/lib/pdfrx.dart
@@ -12,3 +12,6 @@ export 'src/widgets/pdf_widgets.dart';
 export 'src/widgets/scroll_interaction/pdf_viewer_scroll_interaction_delegate.dart';
 export 'src/widgets/scroll_interaction/pdf_viewer_scroll_interaction_delegate_instant.dart';
 export 'src/widgets/scroll_interaction/pdf_viewer_scroll_interaction_delegate_physics.dart';
+export 'src/widgets/sizing/pdf_viewer_size_delegate.dart';
+export 'src/widgets/sizing/pdf_viewer_size_delegate_legacy.dart';
+export 'src/widgets/sizing/pdf_viewer_size_delegate_smart.dart';
--- a/packages/pdfrx/lib/src/widgets/pdf_viewer.dart
+++ b/packages/pdfrx/lib/src/widgets/pdf_viewer.dart
@@ -221,10 +221,7 @@ class _PdfViewerState extends State<PdfViewer>
   PdfDocument? _document;
   PdfPageLayout? _layout;
   Size? _viewSize;
-  double? _coverScale;
-  double? _alternativeFitScale;
-  static const _defaultMinScale = 0.1;
-  double _minScale = _defaultMinScale;
+  late PdfViewerLayoutMetrics _layoutMetrics;
   int? _pageNumber;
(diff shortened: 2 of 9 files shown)


quaaantumdev:
accidentially used the wrong branch for this pull request, there now is a new pull request replacing this one, see https://github.com/espresso3389/pdfrx/pull/581

PR #582, opened 2026-01-02
Improved positioning and scaling with a pluggable size delegate

quaaantumdev:
**Note to Reviewers: I have a version of this that builds on on top of https://github.com/espresso3389/pdfrx/pull/581, so I'd prefer merging https://github.com/espresso3389/pdfrx/pull/581 before this one so I can provide the necessary updates to where sizing/scrolling interacts. But it's also okey to review or merge this one independently, it will just need some manual merging and i have that ready the other way round.**

Currently, resizing the `PdfViewer` (e.g., resizing a browser window, rotating a mobile device, or adjusting split-screen views) results in a rigid experience. The content stays anchored to the top-left corner, meaning that shrinking the window clips the content on the right/bottom, and enlarging the window adds empty whitespace. It lacks the responsive, "responsive web design" feel where content stays centered and adapts to the new container size.

We wanted to enable a modern, adaptive resizing behavior—keeping content centered and automatically scaling to fit the width—without breaking the strict behavior existing users rely on.

### The Solution: Pluggable Sizing Architecture
We refactored the layout initialization and update logic out of `_PdfViewerState` into a strategy pattern, similar to our previous work on `InteractionDelegate`.

*   **`PdfViewerSizeDelegate`**: An abstract base class that defines how the viewer calculates the initial zoom, determines min/max scale limits, and reacts to layout/size changes.
*   **`PdfViewerSizeDelegateProvider`**: A factory pattern used in `PdfViewerParams` to ensure stable lifecycle management across widget rebuilds.

We implemented two concrete versions:

1.  **`...ProviderLegacy` (Default)**: Replicates the exact behavior of the current version; we moved the existing code into this class. It maintains strict top-left alignment and preserves absolute zoom levels during resize. **This ensures no breaking changes for existing users.**
2.  **`...ProviderSmart` (New)**: A responsive strategy designed for Desktop/Web.
    *   **Centering:** Keeps content centered in the viewport during resize.
    *   **Adaptive Scaling:** If the window shrinks, the content scales down to fit ("catches on"). If the window grows, it scales up to fit the width (up to a configurable limit).
    *   **Smart Init:** Defaults to "Fit Width" instead of "Fit Page," which provides a much better initial experience on large monitors.

### Additional Architectural Improvements
While refactoring the sizing logic, we also decoupled the zoom interaction logic:

*   **`PdfViewerZoomStepsDelegate`**: We extracted the logic that calculates "zoom stops" (snap points used for double-tap zooming) into its own delegate.
[...]

Issue #112, opened 2024-04-03
Snap pages?


mobiledevh:
Hello,

First of all, thank you for this awesome lib, best one yet for viewing PDF files with support for overlay widgets on multiple pages, that persists with zoom etc... excellent :)

I only have one small thing remaining, page snapping...
I can't find a way to "snap" pages, meaning the page snaps back to fit the view if the user started scrolling to a page. (ie. you can't have a scroll that stops between 2 pages, it automatically snap to the pages we see the most).

I don't think it's possible through layoutPages so I don't really know where to look :/

Thanks a lot!

musaffa:
@espresso3389 We need this feature badly. How can we help? Do you think [PageView](https://api.flutter.dev/flutter/widgets/PageView-class.html) might help?

burekas7:
@espresso3389 
+1

zhangxiang-1222:
+ 1

espresso3389:
It is fixed if we merge #505, but it has several breaking changes; the issue will be fixed on pdfrx 3.0 (#534).

davidengler:
Yep this is really needed, thank you

PR #589, opened 2026-01-10
Layout enhancements discrete transitions

enhancient:
Resubmission of original PR: https://github.com/espresso3389/pdfrx/pull/505


espresso3389:
I've finally merged PR #581 and #582 and they will be released as 2.3.0 because they does not break the compatibility with existing 2.X series.

2.3.0 contains the following changes:

```
- NEW: Pluggable scroll/zoom interaction architecture ([#581](https://github.com/espresso3389/pdfrx/pull/581))
  - `PdfViewerScrollInteractionDelegateProviderInstant` (default) - instant updates (legacy behavior)
  - `PdfViewerScrollInteractionDelegateProviderPhysics` - smooth, physics-based animations for mouse wheel and trackpad
  - New parameter `scaleByPointerScale` to control trackpad pinch/Ctrl+scroll sensitivity
  - Shift+scroll now triggers horizontal scrolling (standard desktop behavior)
- NEW: Pluggable sizing/layout architecture ([#582](https://github.com/espresso3389/pdfrx/pull/582))
  - `PdfViewerSizeDelegateProviderLegacy` (default) - maintains existing behavior
  - `PdfViewerSizeDelegateProviderSmart` - responsive resizing with content centering and adaptive scaling
  - `PdfViewerZoomStepsDelegate` for customizable double-tap zoom snap points
- DEPRECATED: `maxScale`, `minScale`, `useAlternativeFitScaleAsMinScale`, `onePassRenderingScaleThreshold`, and `calculateInitialZoom` parameters in `PdfViewerParams` - use `sizeDelegateProvider` instead
- NEW: `PdfViewerController.maxScale` getter
- NEW: `PdfViewerController.goToPosition()` method
```

The changes are really conflicting with your PRs because they also change the pdfrx'z zooming logic but it does not break consistency with the existing versions (and it's the reason why I want to release it as 2.X series).

I'm really sorry to delay merging your PRs but I want to finalize pdfrx 2.X series before moving toward the pdfrx 3.X.

Anyway, the current master is ready to accept your PR at last. Please forgive me to bother you so many times before merging your PR.

enhancient:
Having analyzed #581 and #582, I believe the smooth scrolling in #581 could be achieved more simply with wheel delta accumulation rather than a parallel animation system (one in InteractiveViewer and another in pdfrx). The architecture for physics based animations is already in InteractiveViewer. 

The sizing features in #582 (centering, adaptive scaling) can be integrated into #589's layout classes.
The layout class architecture in #589 is designed to be extensible—we could add a centerOnResize option or similar without needing the full SizeDelegate pattern.

Given this PR provides the basis for more complex page layout features and discrete page transitions, I feel like we should try and work in the features of #581 and #582 to this PR rather than the other way round.

espresso3389:
> Given this PR provides the basis for more complex page layout features and discrete page transitions, I feel like we should try and work in the features of https://github.com/espresso3389/pdfrx/pull/581 and https://github.com/espresso3389/pdfrx/pull/582 to this PR rather than the other way round.

~~If you think so, I think it's better to re-implement everything on your PR and I will revert the  merges.~~

Oh, I remember the reason.
The major reason why I merged the PRs is they does not break the compatibility.
Your one is better in the feature but it introduces several imcompatibilities... Mmm.

If I release 2.3.0 as it is before switching to 3.0.0, pdfrx introduces two different interfaces in very short term... So, the best way is anyway, just to abandon to release the PRs.....??

quaaantumdev:
> Having analyzed #581 and #582, I believe the smooth scrolling in #581 could be achieved more simply with wheel delta accumulation rather than a parallel animation system (one in InteractiveViewer and another in pdfrx). The architecture for physics based animations is already in InteractiveViewer.

`InteractiveViewer` currently is a fork of flutter's `InteractiveViewer` with minimal modifications for some callbacks, right? It just seems like the better approach to separate things out into dedicated classes (separation of concerns) instead of packing even more spaghetti code in there, making future merges even harder. 

Your feature very nice and i think it can play fairly well with what came in https://github.com/espresso3389/pdfrx/pull/581 and https://github.com/espresso3389/pdfrx/pull/582. I might be able to take a look at your code and see if i can come up with a clean way to integrate it, if you want me to.

Overall what happens at a given point with user interaction will always be an opinionated choice. Gestures inherently fight each at some point. So i think having a fairly clean classes to integrate and configure these behaviors seems is a worthwhile effort.

espresso3389:
@quaaantumdev I like your Provider approach and it's promising; it effectively removes some of ancient and bad zooming parameters; they are planned to be removed on pdfrx 3.0 anyway.

And, also, I want to merge "the features" introduced by this PR #589... But, frankly speaking, I'm not good at merging such a large changes and I could not find a way to deal with the current situation...

I want to hear voices from both of you (@quaaantumdev and @enhancient) to determine how to do that.

quaaantumdev:
Hi @espresso3389 and @enhancient,

I’ve been looking into this a bit (this PR and the original https://github.com/espresso3389/pdfrx/pull/505). And i think we should discuss the architecture before we settle on the implementation. I think this feature is great but it should be integrated much cleaner and thereby keep the package easier to maintain. Let me explain what I think I've understood so far:

### 1. Dissecting the "Discrete" behavior
If we strip away the term "Discrete Mode," this PR is actually trying to implement three distinct behaviors that are currently being conflated inside the viewer logic:

1.  **Axis Locking:** In standard mode, you can pan diagonally. In this new mode, if the user swipes horizontally, the logic needs to *lock* vertical movement (and vice versa) to simulate a slider or e-book reader experience.
2.  **Intent Interpretation:** The viewer needs to decide contextually: is this drag meant to **Pan** (move around a zoomed-in page) or **Turn** (flip to the next page)? This usually depends on the current zoom level (fit-to-screen vs. zoomed-in).
3.  **Targeted Snapping:** Unlike the standard viewer which uses physics (inertia/friction) to stop wherever it lands, this mode needs to override physics. It must calculate a specific target coordinate (the start of a page) based on a threshold and force the view to animate exactly there.

### 2. The Architectural Issue
Currently, `pdfrx` maintains a copy of Flutter's `InteractiveViewer`. This component should ideally remain "dumb"—it should simply receive matrix transformations (Zoom/Pan) and apply them. It shouldn't know what a "Page" is, nor should it contain logic for specific reading modes.

Implementing "Discrete Mode" by injecting many conditional checks directly into the `InteractiveViewer`'s gesture detection loops adds technical debt. It couples the low-level rendering engine with high-level application logic and will make it very hard to maintain.

### 3. The Proposed Solution: A Touch Interaction Delegate
We have recently merged PR https://github.com/espresso3389/pdfrx/pull/582 (which introduced the `PdfViewerSizeDelegate` to handle *layout and geometry*) and https://github.com/espresso3389/pdfrx/pull/581 (which introduced the `PdfViewerScrollInteractionDelegate` to handle *mouse and trackpad* inputs).

We are missing the third piece of the puzzle: a **`PdfViewerTouchInteractionDelegate`** to handle touch interactions.

The logic in this PR shouldn't live in the core viewer; it should be an isolated strategy that plugs into this new delegate slot. This would allow us to switch behaviors cleanly:

**Strategy A: The Standard "Free Roam" Delegate**
*   **On Drag:** Updates the viewport matrix 1:1.
*   **On Release:** Applies standard inertia/friction to slide to a stop naturally.

**Strategy B: The "Discrete" Delegate (This PR's logic)**
*   **On Drag:** The delegate checks the zoom level. If the document is fit-to-screen, it overrides the standard behavior to lock the axis and treats the drag as a "page turn progress" indicator.
*   **On Release:** Instead of letting physics take over, the delegate intercepts the end of the gesture. It calculates the velocity and distance, asks the **Size Delegate's layout snapshot** for the exact coordinates of the next/previous page, and commands the viewer to animate directly to that target.

### Why this approach?
This separates concerns perfectly. The `InteractiveViewer` handles raw touch events. The `SizeDelegate` knows where the pages are. The `TouchDelegate` decides how the user moves between them.

This allows us to implement the "Discrete" features now, but leaves the door open for other future behaviors (like "Snap-to-Grid" or "Vertical-Only Scrolling") without ever needing to modify the core viewer code again.

I am happy to help outline this abstraction, as it follows the exact patterns we just established in https://github.com/espresso3389/pdfrx/pull/581 and https://github.com/espresso3389/pdfrx/pull/582.

enhancient:
Hi @quaaantumdev and @espresso3389,

Thanks for the analysis, and I agree with these behaviours identified, although axis locking is not specific to discrete mode - if the page is zoomed out to fit the content is should lock on the axis.

I'd push back on your architectural conclusions though  - the boundary-based approach in #589 doesn't inject page knowledge into InteractiveViewer—it calculates boundaries externally and passes them through the existing boundaryMargin API. Flutter's own PageView achieves discrete snapping through PageScrollPhysics + boundary configuration, not through a delegate that intercepts touch events. #589 follows the same pattern.  You could argue that the implementation in #589 is a hybrid approach, where snapping logic is in pdfrx and boundaries are in InteractiveViewer, and we could look to completely abstract the snapping into a ScrollPhysics - I'm open to looking into this.

And to answer your previous question about the fork of InteractiveViewer in pdfrx - it's more than a few callbacks -  it includes physics based scrolling and zooming functionality.  InteractiveViewer was originally designed for panning of images and did not intend to have applications like PDF viewing in mind, so did not including ScrollPhysics like Scrollables. If CustomScrollView had supported both horizontal and vertical scrolling *and* zooming I'm sure @espresso3389  would have used it, but alas InteractiveViewer is the closest flutter has. This PR #477 added ScrollPhysics to InteractiveViewer addiing the expected native scrolling behaviour  for various platforms. 

Boundaries are passed through to InteractiveViewer to enable the ScrollPhysics to bounce back on iOS/MacOS platforms. And actually several popular commercial PDF viewers on Android also typically mimmic this behavior, so we added a FixedOverScrollPhysics to replicate this on Android. In discrete mode passing boundaries for the individual page allows a zoomed in page to likewise bounce off these boundaries - https://github.com/user-attachments/assets/c39e8fc8-64f8-4843-a85e-a52ed5accd53.

I spent several months back and forth with @espresso3389 and weeks of development time on #505/#589, so naturally I'm concerned about completely refactoring this to fit a new architectural approach. People are waiting on discrete pages (https://github.com/espresso3389/pdfrx/issues/112 is the top upvoted issue), #589 addresses this, and @espresso3389 wants to merge this as part of pdfrx 3.0, and I'm keen to support this. It's exciting to see another enthusiastic collaborator in the mix, and we all want pdfrx to be awesome, but I think @espresso3389 needs to make a call on how we move forward here.

espresso3389:
@enhancient Thanks for your comment. I just remembered some of the concepts pdfrx adopted in the past.

@enhancien, @quaaantumdev 
When I started developing pdfrx, I though I could not maintain InteractiveViewer myself and I kept it unchanged as much as possible. But, now, I think, our team (I mean, @enhancient, @quaaantumdev, and me) can maintain the changes to some extent though we'd better keep the changes small as much as possible.

At this moment, I think we'd better release the current master as pdfrx 2.3.0 anyway. It it the "final" version on pdfrx 2 unless there are some critical issues on it.

And, then, we'd make PR #589 feasible to merge onto the master; it will be 3.0 and we can do relatively aggressive modifications on it; we can freely discuss how the changes should be.

enhancient:
@espresso3389  InteractiveViewer doesn't change much in Flutter over many years so I agree it's safe to maintain the fork.

As for making PR #589 feasible to merge onto master - this is not a simple merge. The conflict is architectural.  My advice is, if the functionality in #581 and #582 is more important than #589 to the project, run with that, but if not, revert these (or at least #582) and start with #589 in pdfrx 3.0 as originally planned and go from there. I'm hesitant to further refactor my work because I'm concerned about significant lost effort.

If 'aggressive modifications' means reverting #581/#582 to enable #589, then it may not make sense to release 2.3.0 with them included.

espresso3389:
> If 'aggressive modifications' means reverting https://github.com/espresso3389/pdfrx/pull/581/https://github.com/espresso3389/pdfrx/pull/582 to enable https://github.com/espresso3389/pdfrx/pull/589, then it may not make sense to release 2.3.0 with them included.

I mean, on 2.X series, I don't want to do aggressive modifications.

@enhancient OK, I understood you are hesitant  to merge #589 on the current master. Actually, I also hesitant to do that because it's too risky. We have only to revert #581 and #582 for now. They are postponed to 3.0 or later.

We still have several issues on #589. It breaks several things. We should fix them and also remove several ancient things just kept for compatibility. Many task ahead before pdfrx 3.0.

Anyway, I'll stop accepting any large PRs until finalizing 3.0 release. No more new features on 2.0 series.

@quaaantumdev Sorry for the conclusion. the PR is so important.

espresso3389:
OK, we're backed to the master commit that merges #505. The recent PR merges (#581, #582, and #592) are abandoned. And, close the PR anyway.

enhancient:
@espresso3389 This all seems like a good approach to me. Look forward to working on pdfrx 3.0.

enhancient:
@espresso3389 Is this still the path forward you're thinking for pdfrx-3.0? Looks like recent commit (64d5714) on master isn't taking this path?

PR #592, opened 2026-01-16
Adding a smarter sizing strategy

quaaantumdev:
@espresso3389 thank you for merging https://github.com/espresso3389/pdfrx/pull/581 and https://github.com/espresso3389/pdfrx/pull/582!

This pull request here:
- adds the missing `PdfViewerSizeDelegateProviderSmart` sizing strategy, this is what you were missing in your comment https://github.com/espresso3389/pdfrx/pull/582#issuecomment-3756850071
- and it switches the interaction delegate to use the layout metrics for minScale/maxScale, which is what you effectively asked for in https://github.com/espresso3389/pdfrx/pull/582#issuecomment-3756733756. This did not come "automatically" with the merge as both were based on the same commit. Now they can build on each other, which was the idea in the first place.

After merging this pull request, both https://github.com/espresso3389/pdfrx/pull/581 and https://github.com/espresso3389/pdfrx/pull/582 are "properly" merged.

Thank you for all your efforts on this great package!

### DIFF
--- a/packages/pdfrx/lib/pdfrx.dart
+++ b/packages/pdfrx/lib/pdfrx.dart
@@ -15,6 +15,7 @@ export 'src/widgets/scroll_interaction/pdf_viewer_scroll_interaction_delegate_in
 export 'src/widgets/scroll_interaction/pdf_viewer_scroll_interaction_delegate_physics.dart';
 export 'src/widgets/sizing/pdf_viewer_size_delegate.dart';
 export 'src/widgets/sizing/pdf_viewer_size_delegate_legacy.dart';
+export 'src/widgets/sizing/pdf_viewer_size_delegate_smart.dart';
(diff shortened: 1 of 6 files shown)


espresso3389:
@quaaantumdev 

BTW,

I found the code like the following one on your code:

```dart
class PdfViewerScrollInteractionDelegateProviderInstant extends PdfViewerScrollInteractionDelegateProvider {
  const PdfViewerScrollInteractionDelegateProviderInstant();

  @override
  PdfViewerScrollInteractionDelegate create() => _PdfViewerScrollInteractionDelegateInstant();

  @override
  bool operator ==(Object other) => other is PdfViewerScrollInteractionDelegateProviderInstant;

  @override
  int get hashCode => runtimeType.hashCode;
}
```

I'm not sure whether it is a common idiom to implement singleton like constant. I understand it's working well and actually no concern on its behavior. Is it your invention or any other reference for such style of implementation?

PR #622, opened 2026-03-22
Fix openData deadlock: use FPDF_LoadMemDocument for in-memory data

jam7:
Hi,

Thank you for developing pdfrx.  It's nice to use not blocking library like this from flutter.

I have one problem.  My app is using openData after loading entire pdf, but it hangs in the middle of openData often if I open >10MB pdf.

I notice openData receives a Uint8List that is already fully in memory. However, when the fileSize exceeds maxSizeToCacheOnMemory (default 1MB), openCustom falls through to FPDF_LoadCustomDocument which uses PdfiumFileAccess with condition variables. This path deadlocks after repeated open/close cycles.

Since the data is already in memory, I modify the code always use FPDF_LoadMemDocument by setting maxSizeToCacheOnMemory to data.length when not specified.

After this modification, openData is very stable for me.  Please consider to use this patch or disable condition variable path in openData.  Thank you!

### DIFF
--- a/packages/pdfrx_engine/lib/src/native/pdfrx_pdfium.dart
+++ b/packages/pdfrx_engine/lib/src/native/pdfrx_pdfium.dart
@@ -295,6 +295,10 @@ class PdfrxEntryFunctionsImpl implements PdfrxEntryFunctions {
     required int? maxSizeToCacheOnMemory,
     required void Function()? onDispose,
   }) {
+    // Data is already in memory; always use FPDF_LoadMemDocument to avoid
+    // PdfiumFileAccess callback path which deadlocks on repeated open/close.
+    maxSizeToCacheOnMemory ??= data.length;
+
     return openCustom(
       read: (buffer, position, size) {
         if (position + size > data.length) {


ojpro:
hey @jam7 

I have faced similar issues, and i don't think setting the maxSizeToCacheOnMemory is a good idea, 
as in my case I have some books with +200mb which will be all saved on the memory and could cause crash.

My solution was to implement a Native Implementation and pass the result using Flutter Channels:
```dart
final pdfFile = await PdfDocument.openCustom(
      read: (Uint8List buffer, int position, int size) async {
        final bytes = await _ioService.readFileBytesAt(
          path,
          position: position,
          size: size,
        );
        if (bytes == null) return 0;

        buffer.setRange(0, bytes.length, bytes);
        return bytes.length;
      },
``` 

> I am working on this [plugin](https://github.com/ojprolab/flutter_saf.git) for IOS and ANDROID that offer more functionalites for IO opteration

jam7:
Hi @ojpro,

Thank you.  That's true if I read +200 MB pdf files.  However, the use case of this `_openData` is different, I guess.  This `_openData`  receives whole data in Uint8List data argument.  This function just read whole data using 1MB chunk and hangs.

At first, I tried to modify caller of this `_openData` to adjust a `maxSizeToCacheOnMemory` argument by caller.  However, this is not an argument caller can be passed (not public).

Even if we pass whole data, using 1MB chunk might be better.  I am not sure about it and make this pull request with an assumption like it doesn't matter.  Sorry if it matter.

jam7:
I've realized that using openData I/F was bad choice.

- If I have data on local storage, I should use openFile instead of openData.
- If I was receiving data from a stream, I should use openCustom like what @ojpro explained.

I've changed my code to use openFile this time.  It solves my problem without modifying the source code.  So, I'm closing this even though the openData still has a problem.  Thank you.

PR #623, opened 2026-03-24
Fix crash in loadStructuredText when text starts with newline

punzar:
## Summary

- `PdfTextFormatter.format()` crashes with `StateError: Bad state: No element` when PDF text begins with a newline character
- When the first character is a newline, `outputCharRects` is still empty at the point where the code accesses `.last`
- Added `isNotEmpty` guard to prevent the crash

## Reproduction

Load a PDF page where the structured text content starts with a `\n` character and call `loadStructuredText()`.

## Fix

Added `outputCharRects.isNotEmpty` check before accessing `outputCharRects.last` in the newline handling branch of `PdfTextFormatter.format()`.


espresso3389:
Sorry, recently pdfrx's master was rollbacked so that I just rebased your PR on the current master.

PR #624, opened 2026-03-24
Fix stream closed crash on document dispose race condition

punzar:
## Summary

- `subject.add()` throws `StateError: Cannot add event after closing` when a PDF document is disposed while background operations are still running
- This is a race condition between `dispose()` (which closes the stream) and async operations like page loading and font checking that try to emit events after disposal
- Added `isDisposed` guards before all `subject.add()` calls in three locations

## Affected methods

- `_notifyMissingFonts()` — can fire after async `_getAndClearMissingFonts()` completes post-dispose
- `_notifyDocumentLoadComplete()` — can be called after progressive page loading completes post-dispose
- `pages` setter — can emit `PdfDocumentPageStatusChangedEvent` after dispose

## Fix

Added `!isDisposed` checks before each `subject.add()` call to silently skip event emission when the document has already been disposed.


espresso3389:
On the current master, same fix was already applied independently.

PR #648, opened 2026-05-24
Fix PDFium build hook output path per platform

adil192:
Fixes an issue where the Linux `libpdfium.so` and the Android `libpdfium.so` can get mixed up:
If `libpdfium.so` exists from a previous Linux build, Android builds also use it which leads to incompatibility and Google Play warnings.
This PR separates the outputFile path by platform to avoid these conflicts.


espresso3389:
Released in pdfium_dart v0.2.3: https://pub.dev/packages/pdfium_dart/versions/0.2.3

This fixes PDFium native asset output paths so platform-specific libraries with the same CPU architecture do not collide.

Written by Codex

PR #649, opened 2026-05-26
Add macOS support to loader

iadept:


espresso3389:
In normal case, Dart native asset resolves `libpdfium.dylib`. Do you need non-native asset case on macOS?

espresso3389:
Because [`getPdfium` internally calls `_getModule`](https://github.com/iadept/pdfrx/blob/2165616eca935ac100d82b92c7605f3bc3dd5f65/packages/pdfium_dart/lib/src/pdfium_loader.dart#L29), if `_getModule` unconditionally returns `libpdfium.dylib` on macOS, it breaks the consistency on codes that depends on the current behavior. If we move [the line](https://github.com/iadept/pdfrx/blob/2165616eca935ac100d82b92c7605f3bc3dd5f65/packages/pdfium_dart/lib/src/pdfium_loader.dart#L39C3-L39C66) into `try-catch`, it may keep the current behavior but...

PR #654, opened 2026-06-12
Fix inverted guard in selectAllText() causing null dereference

koji-1009:
I wanted to use pdfrx, so I ran a code review on it and found a minor bug.
It is rare in practice but can happen, and the fix is small, so I'd like to fix it.

`_layout` is null for a short time after a document is loaded, before its layout is ready. In this time, `selectAllText()` can be called by Cmd/Ctrl+A. The old guard condition was wrong, so it did not return and crashed on `_layout!`.


espresso3389:
The fix for this PR has been released in [pdfrx v2.4.4](https://pub.dev/packages/pdfrx/versions/2.4.4).

Fix summary: PdfViewer.selectAllText() no longer attempts text selection before layout is available, avoiding a null-check crash.

Written by Codex

PR #655, opened 2026-06-13
Fix PDFium initialization flag set in the wrong isolate

koji-1009:
The old code set `_initialized = true` inside `BackgroundWorker.compute()`, which runs in a worker isolate, so the flag was never set on the caller isolate (top-level variables are not shared between isolates). The init guard therefore never triggered, and PDFium was initialized again on every call. The call was also not awaited.

This PR sets the flag on the caller isolate (after awaiting the worker), and adds a test for it.

### DIFF
--- a/packages/pdfrx_engine/lib/src/native/pdfrx_pdfium.dart
+++ b/packages/pdfrx_engine/lib/src/native/pdfrx_pdfium.dart
@@ -9,6 +9,7 @@ import 'dart:typed_data';
 
 import 'package:crypto/crypto.dart';
 import 'package:ffi/ffi.dart';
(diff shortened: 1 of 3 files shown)


espresso3389:
The fix for this PR has been released in [pdfrx v2.4.4](https://pub.dev/packages/pdfrx/versions/2.4.4) and [pdfrx_engine v0.4.3](https://pub.dev/packages/pdfrx_engine/versions/0.4.3).

Fix summary: PDFium initialization is now tracked on the caller isolate, avoiding repeated initialization attempts after worker-isolate initialization.

Written by Codex

PR #673, opened 2026-07-12
Enable Windows ARM64 build

kyakuno:
Uncomment the architecture detection so that pdfium-win-arm64 is downloaded when building for ARM64. The referenced pdfium-binaries release (chromium/7202) ships win-arm64 binaries.


espresso3389:
If you apply the patch to the current pdfrx, can you build & run the resulting executable on Windows ARM64 using the stable Flutter toolchain?

I don't have ARM64 Windows on my side and could not test it directly. The current build test on GitHub Actions still uses Flutter mater for Windows ARM64 but can we switch to the stable?

espresso3389:
And, anyway, I could not do anything with the PR:

image

kyakuno:
@espresso3389 
Thank you for confirming.
In the 1.x series, pdfrx used the CMake format and needed Arm64 support, but it appeared that the latest main has switched to the native asset format.
As a result, by upgrading to 2.4.5, Arm64 support was possible without needing to fork it. Therefore, I will withdraw this PR.
Also, with 2.4.5, it worked normally on Windows on Arm as well.

PR #693, opened 2026-08-12
fix: BackgroundWorker.compute hangs forever on a genuinely async callback

pandeyshivam10:
## Problem

`BackgroundWorker.compute()` hangs forever (never resolves, no error surfaced) when the callback passed to it does genuine asynchronous work — i.e. actually awaits something instead of completing synchronously.

## Root cause

`_ExecuteParams.execute()` ran:

```dart
sendPort.send(callback(message));
```

`callback` returns `FutureOr<R>`. When it happens to complete synchronously, this works by accident, since the value is already a plain `R`. But when it returns a *pending* `Future<R>` (real async work), `SendPort.send()` throws:

```
Invalid argument(s): Illegal argument in isolate message: object is unsendable - Library:'dart:async' Class: _Future
```

A pending `Future` is not a sendable isolate message. This throw happens inside the worker isolate, where it's swallowed — the caller's `await receivePort.first` just waits forever with no error ever crossing back.

A related gap: a *synchronous* throw inside the callback also had no error-forwarding path, so that case hung too.

## Fix

`execute()` now always awaits the callback (via an `async` closure, so sync and async callbacks are handled the same way) and sends back one of two plain, always-sendable wrapper objects instead of the raw result:

- `_ComputeResult<R>(value)` on success
- `_ComputeError(errorString, stackTraceString)` on failure — the original error/stack trace objects aren't guaranteed sendable either (arbitrary exception types can hold non-sendable fields), so only their string forms cross the isolate boundary

The caller side (`_compute()`) checks for `_ComputeError` and rethrows a `_WorkerComputeException` with the reconstructed stack trace; otherwise it unwraps `_ComputeResult.value`.

## Testing

Added `test/background_worker_test.dart` covering:
- the original synchronous case (no regression)
- a callback that does genuine async work (real `await`)
- a callback with multiple sequential `await`s
- a callback that throws synchronously
- a callback that throws asynchronously (after an `await`)

All pass. `dart analyze` is clean, no regressions in the existing suite.

### DIFF
--- a/packages/pdfrx_engine/lib/src/native/worker.dart
+++ b/packages/pdfrx_engine/lib/src/native/worker.dart
@@ -109,7 +109,17 @@ class BackgroundWorker {
   /// Inside [callback], you can only use passed message and create new objects.
   /// You cannot access any variables from the outer scope, otherwise, it will throw an error.
   Future<R> _compute<M, R>(PdfrxComputeCallback<M, R> callback, M message) async {
-    return await _sendComputeParams((sendPort) => _ExecuteParams(sendPort, callback, message)) as R;
+    final result = await _sendComputeParams((sendPort) => _ExecuteParams(sendPort, callback, message));
+    if (result is _ComputeError) {
+      // The original error/stack trace object may itself be unsendable (arbitrary user exception types can hold
+      // non-sendable fields), so only their string forms cross the isolate boundary; reconstruct a generic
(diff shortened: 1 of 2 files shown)


espresso3389:
This fix has been released in [pdfrx_engine 0.4.7](https://pub.dev/packages/pdfrx_engine/versions/0.4.7).

Written by Codex

PR #695, opened 2026-08-19
Keep text selection alive during progressive page loading

Guesub:
### Problem

`_onDocumentEvent` unconditionally calls `_clearTextSelections` on every `PdfDocumentPageStatusChangedEvent`. During progressive loading of a large document these events arrive continuously (one per load batch, every ~250 ms), so any text selection the user makes is wiped almost immediately — on a 47,352-page document the selection context menu appears and vanishes within a fraction of a second, and text selection/copy is impossible until the whole document finishes loading.

### Fix

Clear the selection only when one of the changed pages actually falls inside the selected page range. Loading events for unrelated (typically far-away, not-yet-loaded) pages no longer destroy the user's selection, while a genuine change to a selected page still invalidates it as before.

Verified on an iPhone with a 1.9 GB / 47,352-page PDF: long-press selection and copy now work immediately after opening, while pages keep loading in the background.

### DIFF
--- a/packages/pdfrx/lib/src/widgets/pdf_viewer.dart
+++ b/packages/pdfrx/lib/src/widgets/pdf_viewer.dart
@@ -540,7 +540,15 @@ class _PdfViewerState extends State<PdfViewer>
         _canvasLinkPainter.releaseLinksForPage(change.key);
         _textCache.remove(change.key);
       }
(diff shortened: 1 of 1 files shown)


espresso3389:

Written by Codex

Issue #110, opened 2024-04-02
Memory leak, image is never disposed


jezell:
https://github.com/espresso3389/pdfrx/blob/master/lib/src/widgets/pdf_widgets.dart

_image is not disposed in dispose()
_image is not disposed in _updateImage

espresso3389:
Fixed on 1.0.51 and other branches.

PR #698, opened 2026-08-24
fix(PdfPageView): dispose the previous image before replacing it

### LINKED ISSUE #697: PdfPageView leaks a ui.Image on every re-render (_updateImage never disposes the previous _image)
### Summary

`PdfPageView` orphans a full-resolution `ui.Image` on every re-render. `_PdfPageViewState._updateImage` overwrites `_image` without disposing the previous one, and it has no `mounted` guard on the assignment itself.

This is the second half of #110, which was closed as "Fixed on 1.0.51 and other branches". Only the `dispose()` half was actually fixed (via `_clearCache()`); the `_updateImage` half is still present on `master` today (`d3729a1`) and in the latest release, 2.4.7.

### Where

`packages/pdfrx/lib/src/widgets/pdf_widgets.dart`, `_PdfPageViewState._updateImage`, lines 447-460 on `master` and on tag `pdfrx-v2.4.7`:

```dart
if (pageImage == null) return;
try {
  final newImage = await pageImage.createImage();
  pageImage.dispose();
  _image = newImage;          // <-- previous _image is never disposed
  if (mounted) {
    setState(() {});
  }
} catch (e) {
  developer.log('Error creating image: $e');
  pageImage.dispose();
}
```

Two distinct leaks:

1. **The previous `_image` is never disposed.** The only `_image?.dispose()` in the class is in `_clearCache()` (line 347), which is reached from `dispose()`, from `didUpdateWidget()` on a document/pageNumber/rotation change, and from the document-events listener — never from this re-render path. So any re-render that changes the requested page size (a zoom tier change, a layout change, a `pageSizeCallback` that returns a new size) allocates a new decoded image and drops the old one on the floor.

2. **No `mounted` guard before the assignment.** The `mounted` check only guards `setState`. If the widget is disposed during the `await`, `_clearCache()` has already run and set `_image = null`; line `_image = newImage` then writes a fresh image into a dead `State` that nothing will ever dispose.

### Why this is inconsistent with the rest of the package

`PdfViewer` gets this right in `packages/pdfrx/lib/src/widgets/pdf_viewer.dart` (lines 1760-1764):

```dart
if (img == null || !mounted || cancellationToken.isCanceled) return;
final newImage = _PdfImageWithScale(await img.createImage(), scale);
cache.pageImages[page.pageNumber]?.dispose();   // old image disposed
cache.pageImages[page.pageNumber] = newImage;
```

and the partial-image path at 1847-1852 also checks `!mounted` before creating the image. So the correct pattern already exists in the codebase; `PdfPageView` simply does not follow it.

### Impact

We hit this in a Flutter e-paper app that renders one `PdfPageView` per page with a zoom-dependent `pageSizeCallback`, so a page re-renders on every zoom tier change. On web the accumulated `ui.Image` allocations are a significant contributor to heap exhaustion; once the heap is full, `pdfium_worker.js` fails a 1 MB `malloc` and the viewer dies on a small page PDF.

### Suggested fix

```dart
final newImage = await pageImage.createImage();
pageImage.dispose();
if (!mounted) {
  newImage.dispose();
  return;
}
_image?.dispose();
_image = newImage;
setState(() {});
```

I have this on a branch and will open a PR shortly.

### Version

- pdfrx `master` @ `d3729a1`, and 2.4.7 (identical file, `diff` clean)
- Present unchanged since at least 1.1.35

bschmalb-ksta:
Fixes #697.

`_PdfPageViewState._updateImage` overwrites `_image` without disposing the previous one, so every re-render that changes the requested page size (a zoom tier change, a layout change, a `pageSizeCallback` returning a new size) orphans a full-resolution decoded `ui.Image`. The only `_image?.dispose()` in the class is in `_clearCache()`, which this path never calls.

It also lacked a `mounted` guard on the assignment — the existing check only guarded `setState`. If the widget was disposed during the `await`, `_clearCache()` had already run and nulled `_image`, and the assignment then wrote a fresh image into a dead `State` where nothing would ever dispose it.

This is the second half of #110, which was closed as fixed in 1.0.51; only the `dispose()` half landed.

The fix matches what `PdfViewer` already does in `pdf_viewer.dart` (lines 1760-1764 and 1847-1852): guard on `!mounted`, dispose the old image, then swap.

```dart
final newImage = await pageImage.createImage();
pageImage.dispose();
if (!mounted) {
  newImage.dispose();
  return;
}
_image?.dispose();
_image = newImage;
setState(() {});
```

Kept deliberately minimal — no reformatting, no changelog entry (happy to add one if you'd like). I could not run `dart analyze` against the workspace locally: dependency resolution fails on `meta` (`pdfium_flutter` wants 1.17.0 via the Flutter SDK's `flutter_test`, `pdfrx_engine` wants ^1.18.0). That failure reproduces identically on unmodified `master`, so it is unrelated to this change.

### DIFF
--- a/packages/pdfrx/lib/src/widgets/pdf_widgets.dart
+++ b/packages/pdfrx/lib/src/widgets/pdf_widgets.dart
@@ -450,10 +450,13 @@ class _PdfPageViewState extends State<PdfPageView> {
     try {
       final newImage = await pageImage.createImage();
       pageImage.dispose();
-      _image = newImage;
-      if (mounted) {
-        setState(() {});
+      if (!mounted) {
+        newImage.dispose();
+        return;
       }
+      _image?.dispose();
+      _image = newImage;
+      setState(() {});
     } catch (e) {
       developer.log('Error creating image: $e');
       pageImage.dispose();


espresso3389:

Written by Codex

PR #699, opened 2026-08-24
fix(createImage): surface decode failures instead of hanging forever

### LINKED ISSUE #696: PdfImage.createImage() hangs forever on a decode failure: the Completer has no error path, so the pdfium buffer leaks and the error escapes to the zone
Version: **pdfrx 2.2.24** / pdfrx_engine 0.3.9. All line numbers are from the published pub.dev archives.

## The problem

`PdfImageExt.createImage()` — `lib/src/pdfrx_flutter.dart:97-107`:

```dart
final comp = Completer<Image>();
decodeImageFromPixels(
  pixels,
  width,
  height,
  PixelFormat.bgra8888,
  (image) => comp.complete(image),
  targetWidth: targetWidth,
  targetHeight: targetHeight,
);
return comp.future;
```

The completer is fed **only** by `decodeImageFromPixels`' success callback, and `dart:ui`'s `decodeImageFromPixels` has no error channel at all — its signature takes a single `ImageDecoderCallback`, and every `.then` in its internal chain (`sky_engine/lib/ui/painting.dart`, in `decodeImageFromPixels`) is written without an `onError`.

So when the decode fails — `ImmutableBuffer.fromUint8List`, `instantiateCodec`, or `getNextFrame()` completing with an error — three things happen:

1. `callback` is never invoked, so `comp.complete` never runs and **`comp.future` never completes**. It hangs, it does not throw.
2. The error propagates out of the engine's unawaited `.then` chain and **escapes as an unhandled async error** to the ambient `Zone` / `PlatformDispatcher.onError`. It never reaches pdfrx's caller.
3. Because it hangs rather than throwing, the `try`/`catch` at `lib/src/widgets/pdf_widgets.dart:448-458` is **dead code for this failure mode**:

```dart
try {
  final newImage = await pageImage.createImage();
  pageImage.dispose();
  ...
} catch (e) {
  developer.log('Error creating image: $e');
  pageImage.dispose();
}
```

The `await` on line 449 never returns, so neither `dispose()` runs.

## Why that matters: a native leak and a latched-blank widget

`PdfImage.dispose()` on the pdfium backend is a real `malloc.free` — `pdfrx_engine-0.3.9/lib/src/native/pdfrx_pdfium.dart:1553-1555`:

```dart
void dispose() {
  malloc.free(_buffer);
}
```

Sitting behind a never-returning `await`, it is unreachable, so **`width * height * 4` bytes of malloc'd native memory leak permanently** on every failed decode, along with the pending `Completer`. The same shape appears in the viewer, where `finally { img?.dispose(); }` (`lib/src/widgets/pdf_viewer.dart:1530` and `:1602`) is likewise inside an `async` body awaiting the same future.

Separately, `_updateImage` commits `_pageSize` *before* rendering and never rolls it back:

```dart
if (pageSize == _pageSize) return;   // pdf_widgets.dart:436
_pageSize = pageSize;                // :437
```

After a failure, every later `build` → `_updateImage` returns early at :436. `_image` stays `null` with no retry and no error state, so the page is **permanently blank** until `_clearCache()` happens to run (document/pageNumber/rotation change, dispose, or a resize to different constraints).

## How we hit it

A Flutter app rendering newspaper pages on low-memory Android devices. Impeller's `DecompressTexture` has to do a BGRA→RGBA conversion, which allocates a second full-size bitmap; when that allocation fails the engine produces `Exception: Could not allocate intermediate for pixel conversion.`

For us that surfaced as ~1,961 Crashlytics events across 149 users, reported as **fatal** — because the only place the error appears is the global `PlatformDispatcher.onError` handler, with no type and no context tying it to an image decode. Users saw a permanently blank page. We could not catch it at any pdfrx call site.

## Suggested fix

Give the completer an error path. `decodeImageFromPixels` cannot report errors, but the pieces it wraps can — building on `ImmutableBuffer`/`ImageDescriptor`/`Codec` directly lets the failure be forwarded:

```dart
final comp = Completer<Image>();
ImmutableBuffer.fromUint8List(pixels).then((buffer) async {
  try {
    final descriptor = ImageDescriptor.raw(
      buffer, width: width, height: height, pixelFormat: PixelFormat.bgra8888,
    );
    final codec = await descriptor.instantiateCodec(
      targetWidth: targetWidth, targetHeight: targetHeight,
    );
    final frame = await codec.getNextFrame();
    comp.complete(frame.image);
  } finally {
    buffer.dispose();
  }
}).onError(comp.completeError);
return comp.future;
```

Any shape that guarantees `comp` always completes would fix all three symptoms at once: the existing `catch` blocks start working, `dispose()` becomes reachable so the native buffer is freed, and callers can finally distinguish "this page failed to decode" from an anonymous zone error.

Resetting `_pageSize` in the failure path of `_updateImage` would additionally let a failed page recover on the next build instead of latching blank.

Happy to open a PR if the approach looks right to you.

bschmalb-ksta:
Fixes #696.

## The problem

`PdfImageExt.createImage()` creates a `Completer<Image>` and hands `decodeImageFromPixels()` a success-only callback:

```dart
final comp = Completer<Image>();
decodeImageFromPixels(
  pixels, width, height, PixelFormat.bgra8888,
  (image) => comp.complete(image),
  targetWidth: targetWidth, targetHeight: targetHeight,
);
return comp.future;
```

`ImageDecoderCallback` has no error parameter, and `decodeImageFromPixels` chains `.then()` without a `.catchError`. So when the engine fails — in practice `Could not allocate intermediate for pixel conversion`, i.e. a large page on a memory-constrained device — the completer is never completed **at all**. The caller does not get an exception, it gets a future that never returns.

Three consequences:

1. **The `try`/`catch` in `PdfPageView._updateImage()` cannot fire**, because there is nothing to catch. `createImage` isn't `async`, so that `catch` only ever saw a synchronous throw from argument validation.
2. **The pdfium buffer leaks.** `pageImage.dispose()` — which is `malloc.free(_buffer)` — sits after the never-returning `await`. Every failed render leaks a full page bitmap.
3. **The engine's error escapes into the zone** as an unhandled async error with nothing in the stack pointing at pdfrx, which makes it very hard to attribute. This is how I found it: it showed up as an uncatchable crash in our own app's error reporting with no pdfrx frames.

## The fix

Drive the codec directly. This is exactly what `decodeImageFromPixels` does internally (`ImmutableBuffer` → `ImageDescriptor.raw` → `instantiateCodec` → `getNextFrame`), except awaited, so a failure propagates to the caller as an exception. `try`/`finally` releases the buffer, descriptor and codec on both paths; the SDK's version disposes the buffer and descriptor only in its last `.then()`, so both leak if the chain rejects earlier.

Behaviour on success is unchanged: same pixel format, same target size logic, same disposal order as the SDK.

With that fixed, the `catch` in `_updateImage()` becomes reachable, so two things there needed adjusting:

- **`pageImage.dispose()` moved into a `finally`.** It was duplicated across the `try` and the `catch`, and skipped entirely if `createImage()` threw.
- **`_pageSize` is restored on failure.** It is committed *before* the render and guarded by `if (pageSize == _pageSize) return;`, so a page that failed once would never be re-rendered at that size again. Without this, fixing the hang just turns one permanently-blank page into another. It only restores if a newer call hasn't already claimed `_pageSize`, so a superseded render can't stomp a live one.

`PdfViewer`'s two `createImage()` call sites already use `finally`, so they only benefit from the first change.

## Tests

`createImage()` had no test coverage, so I added some: native-size decode, downscaling to `pixelSizeThreshold`, the already-under-threshold no-op, and the regression itself — a `PdfImage` claiming 4000x3000 over a 4x3 buffer. That last one asserts a throw; on `master` it hangs until the suite times out:

```
00:00 +3: throws instead of hanging when the buffer cannot be decoded
Shell: [ERROR:flutter/lib/ui/painting/image_decoder_skia.cc(90)] Could not create image from decompressed bytes.
00:00 +3 -1: throws instead of hanging when the buffer cannot be decoded [E]
  Exception: Codec failed to produce an image, possibly due to invalid image data.

  TimeoutException after 0:00:30.000000: Test timed out after 30 seconds.
```

Note what that shows: the engine *did* report the failure, as `Exception: Codec failed to produce an image, possibly due to invalid image data.` — it just arrived as an unhandled error in the zone rather than at the `await`, which is consequence 3 above. With the fix the same case throws at the call site and the test passes in under a second.

## Checks

`flutter analyze` clean, `dart format` reports no changes, and the new tests pass. Of the pre-existing tests, `pdfium_loading_test.dart` and `pdf_viewer_test.dart` need a real pdfium binary and network access respectively, so I couldn't run them here — happy to adjust if CI flags anything.

No CHANGELOG entry, matching #698. Let me know if you'd prefer one.

### DIFF
--- a/packages/pdfrx/lib/src/pdfrx_flutter.dart
+++ b/packages/pdfrx/lib/src/pdfrx_flutter.dart
@@ -1,4 +1,3 @@
-import 'dart:async';
 import 'dart:ui';
 
 import 'package:flutter/foundation.dart';
@@ -58,7 +57,11 @@ extension PdfImageExt on PdfImage {
   /// while maintaining the aspect ratio.
   ///
   /// The returned [Image] must be disposed of when no longer needed.
-  Future<Image> createImage({int? pixelSizeThreshold}) {
+  ///
+  /// Throws if the engine cannot decode the pixel buffer, which in practice
+  /// means it failed to allocate for it - a large page on a memory-constrained
+  /// device. The returned future never completes with null.
+  Future<Image> createImage({int? pixelSizeThreshold}) async {
     int? targetWidth;
     int? targetHeight;
     if (pixelSizeThreshold != null && (width > pixelSizeThreshold || height > pixelSizeThreshold)) {
@@ -72,17 +75,27 @@ extension PdfImageExt on PdfImage {
       }
     }
 
-    final comp = Completer<Image>();
-    decodeImageFromPixels(
-      pixels,
-      width,
-      height,
-      PixelFormat.bgra8888,
-      (image) => comp.complete(image),
-      targetWidth: targetWidth,
-      targetHeight: targetHeight,
-    );
-    return comp.future;
+    // This is what `decodeImageFromPixels` does internally, driven directly
+    // instead. That function only takes a success callback, so a decode or
+    // allocation failure leaves the caller awaiting a future that never
+    // completes: the page stays blank, the `pixels` buffer is never freed
+    // because the caller's `dispose()` sits after the await, and the error
+    // surfaces as an unhandled zone error with no context. Awaiting the codec
+    // ourselves lets the failure propagate to the caller as an exception.
+    final buffer = await ImmutableBuffer.fromUint8List(pixels);
+    ImageDescriptor? descriptor;
+    try {
+      descriptor = ImageDescriptor.raw(buffer, width: width, height: height, pixelFormat: PixelFormat.bgra8888);
+      final codec = await descriptor.instantiateCodec(targetWidth: targetWidth, targetHeight: targetHeight);
+      try {
+        return (await codec.getNextFrame()).image;
+      } finally {
+        codec.dispose();
+      }
+    } finally {
+      descriptor?.dispose();
+      buffer.dispose();
+    }
   }
 }
 
--- a/packages/pdfrx/lib/src/widgets/pdf_widgets.dart
+++ b/packages/pdfrx/lib/src/widgets/pdf_widgets.dart
@@ -436,6 +436,7 @@ class _PdfPageViewState extends State<PdfPageView> {
     }
 
     if (pageSize == _pageSize) return;
+    final previousPageSize = _pageSize;
     _pageSize = pageSize;
 
     _cancellationToken?.cancel();
@@ -449,7 +450,6 @@ class _PdfPageViewState extends State<PdfPageView> {
     if (pageImage == null) return;
     try {
       final newImage = await pageImage.createImage();
-      pageImage.dispose();
       if (!mounted) {
         newImage.dispose();
         return;
@@ -459,6 +459,11 @@ class _PdfPageViewState extends State<PdfPageView> {
       setState(() {});
     } catch (e) {
       developer.log('Error creating image: $e');
+      // Restore the size we came in with, but only if a newer call has not
+      // already claimed `_pageSize`. Otherwise the equality guard above locks
+      // this page out of ever retrying and it stays blank for good.
+      if (_pageSize == pageSize) _pageSize = previousPageSize;
+    } finally {
       pageImage.dispose();
     }
   }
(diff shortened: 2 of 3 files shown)


bschmalb-ksta:
@espresso3389 you're a machine! TY for the quick merges ❤️

espresso3389:

Written by Codex

PR #704, opened 2026-09-04
feat(pdfium_dart): load PDFium on HarmonyOS (OHOS) via bare-name dlopen

HsukqiLee:
## Summary

Adds HarmonyOS (OpenHarmony Flutter) support to `getPdfium()`'s platform resolution in `pdfium_dart`.

## What it does

`_getModuleFileName()` now returns `libpdfium.so` when `Platform.operatingSystem == 'ohos'`, mirroring the Android bare-name `dlopen` strategy. On OH Flutter the loader resolves the library from the app's native library directory, so apps bundle `libpdfium.so` under the HAP's `libs/<abi>/` directory (documented in the README platform table).

## Compatibility

- Detection uses the `Platform.operatingSystem` string, since standard Dart SDKs expose no `Platform.isOhos`; the check is a no-op on every platform this package currently supports, so there is no behavior change elsewhere.
- The build hook is untouched: it does not target OHOS, and no native asset is produced there — hence the bring-your-own-library note.

## Verification

Verified in a real HarmonyOS NEXT HAP build (OH Flutter 3.44.x canary / API 24): with `libpdfium.so` (ARM64, musl) placed in `entry/libs/arm64-v8a/`, the HAP packages the library and `getPdfium()` resolves it at runtime.

### DIFF
--- a/packages/pdfium_dart/README.md
+++ b/packages/pdfium_dart/README.md
@@ -108,5 +108,8 @@ The bindings are generated from PDFium headers using the `ffigen` configuration 
 | Linux    | x64, ARM64, ARM, x86 | ✅ |
 | Android  | ARM64, ARMv7, x86, x86_64 | ✅ |
 | macOS    | x64, ARM64  | ✅      |
+| HarmonyOS (OHOS) | ARM64 | ✅ (bring your own `libpdfium.so`) |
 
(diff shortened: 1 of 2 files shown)


espresso3389:
Although I'm not sure whether it actually works on HarmonyOS or not, it seems OK and no conflict with other platforms.

PR #708, opened 2026-09-04
feat: run page renders ahead of background work on the PDFium worker

palmoni5:
## Summary

All PDFium work in the process funnels through a single worker isolate (`BackgroundWorker`). An app that extracts text, parses outlines or otherwise indexes a document in the background while a viewer is open makes the renders of the visible pages queue behind that work, which shows up as pages staying blank or blurry while the background job runs.

This adds an internal two-level priority to `BackgroundWorker.compute` / `computeWithArena` (`BackgroundWorkerPriority.high` / `.normal`). `PdfPage.render` and `PdfDocument.reloadPages` use `high`; everything else keeps the default `normal`. Callers that never pass a priority see exactly the same FIFO behaviour as before.

## Ordering guarantee

An item never runs before an item of equal-or-higher priority that was enqueued earlier:

- a `high` item may overtake `normal` items enqueued before it;
- a `normal` item never overtakes a `high` item enqueued before it;
- items of the same priority run in FIFO order.

The second point matters for `PdfDocument.dispose()`: it marks the document disposed synchronously and then queues `FPDF_CloseDocument` at `normal` priority, so renders that already passed the disposed check and are queued as `high` items are guaranteed to run before the close and never touch a freed `FPDF_DOCUMENT`. For that reason there is deliberately no starvation guard that lets normal work through while high work is pending. High priority is only used for rendering and measuring the pages that are currently visible, so a burst of it is bounded by the visible page set and background work simply waits for the burst to end.

## Scheduling detail

- Incoming compute requests are queued per priority in the worker isolate instead of being executed on arrival, and drained one item per event-loop turn via `Timer.run`. Yielding between items is what lets a `high` message that is already sitting in the isolate's mailbox be picked up before the remaining `normal` items. A microtask would not yield, so port messages would not be seen until the queue was empty.
- Every callback is started synchronously inside its turn, so the relative order of same-priority items is unchanged and a worker that only receives `normal` work behaves like the previous plain FIFO.
- `suspendDuringAction` and `stop` drain all queued work synchronously before acknowledging, so their "everything sent earlier has run" guarantee is preserved (for synchronous callbacks, which is what every PDFium callback in the engine is; this is now documented on both).
- The drain is re-armed in a `finally` so a synchronous throw in an item cannot leave the remaining items stuck in the queue.

## Real-app verification

Verified in a Flutter desktop app (Windows) that keeps a viewer open while it pre-parses the outlines of ~40 scanned volumes and extracts text for indexing in the background: the app ran on this branch with no errors, and visible pages rendered while the background work was running.

Engine-level numbers on two scanned volumes (281 and 242 pages, ~21 MB and ~38 MB), rendering the open page at 1200 px width every 100 ms while `loadPagesProgressively` measures the rest of the document, this branch vs. `master` (2.6.1):

| | idle render | render during measurement, p50 | render during measurement, max |
|---|---|---|---|
| 281-page volume, master | 112 ms | 256 ms | 361 ms |
| 281-page volume, this PR | 130 ms | 256 ms | 263 ms |
| 242-page volume, master | 79 ms | 227 ms | 334 ms |
| 242-page volume, this PR | 79 ms | 225 ms | 233 ms |

The p50 is unchanged, as expected: a measurement slice that is already executing cannot be interrupted, and the default slice is 250 ms. The tail is what this PR fixes: on `master` a render could also land behind other queued work, on this branch it waits for at most the currently executing item. Shortening the measurement slices so a render can slot in between pages is a separate, complementary change.

## Testing

- `dart test` in `packages/pdfrx_engine`: 65 passed, including 7 scheduler tests in `background_worker_test.dart` (high overtakes queued normal, FIFO for default calls, `computeWithArena` priority, error propagation for both priorities, suspend/resume buffering and ordering, and a regression test that queues 12 high computes followed by one normal compute and asserts the normal one completes last).
- `dart analyze` in `packages/pdfrx_engine`: only the two pre-existing `prefer_initializing_formals` infos in `pdfrx_pdfium.dart` (present on `master`).
- `flutter analyze` in `packages/pdfrx`: only the pre-existing infos.

### DIFF
--- a/packages/pdfrx_engine/lib/src/native/pdfrx_pdfium.dart
+++ b/packages/pdfrx_engine/lib/src/native/pdfrx_pdfium.dart
@@ -1324,6 +1324,7 @@ class _PdfDocumentPdfium extends PdfDocument {
         return (pages: pages, invalidPageNumber: invalidPageNumber);
       },
       (docAddress: document.address, pageNumbersToReload: pageNumbersToReload, currentPageCount: _pages.length),
+      priority: BackgroundWorkerPriority.high,
     );
     if (results.invalidPageNumber != null) {
       throw ArgumentError.value(
@@ -1764,6 +1765,7 @@ class _PdfPagePdfium extends PdfPage with PdfPageLinkCache {
             formInfo: document.formInfo.address,
             cancelFlag: cancelFlag.address,
           ),
+          priority: BackgroundWorkerPriority.high,
         );
       });
 
--- a/packages/pdfrx_engine/lib/src/native/worker.dart
+++ b/packages/pdfrx_engine/lib/src/native/worker.dart
@@ -10,6 +10,20 @@ import '../pdfrx.dart';
 
 typedef PdfrxComputeCallback<M, R> = FutureOr<R> Function(M message);
 
+/// Scheduling priority of a [BackgroundWorker.compute] call.
+///
+/// All PDFium work in the process funnels through a single worker isolate, so a long run of background work
+/// (text extraction, outline parsing, document loading, ...) can delay latency-sensitive work such as rendering the
+/// pages currently on screen. The worker keeps one queue per priority and always picks a pending [high] item before
+/// any [normal] one; items of the same priority run in FIFO order.
+enum BackgroundWorkerPriority {
+  /// Latency-sensitive work (page rendering, on-demand measurement of visible pages).
+  high,
+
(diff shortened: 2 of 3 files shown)


palmoni5:
@espresso3389 
I’d love to hear if there’s any progress on this; this fix is ​​very important for my app.

PR #709, opened 2026-09-05
perf: measure pages in short worker chunks during progressive loading

palmoni5:
## Summary

`_PdfDocumentPdfium.loadPagesProgressively` sends each `loadUnitDuration` slice (250 ms by default) to the PDFium worker as one `computeWithArena` task that loops `FPDF_LoadPage`/`FPDF_ClosePage` over the unmeasured pages until the slice elapses. All PDFium calls in the process share that single worker, so a `PdfPage.render` of the visible page issued while the rest of the document is being measured sits behind the task and waits for the whole slice. On two scanned volumes (281 and 242 pages, roughly 9-11 ms per page to measure) this showed up as render latency of p50 256 / 227 ms during measurement versus 112 / 79 ms idle, for the first few seconds after opening the document.

This PR keeps the `loadUnitDuration` budget but spends it as a series of short worker tasks (`_measurementChunkDuration`, 20 ms: about two pages on a typical desktop, small enough that a queued render waits roughly one page, large enough that the per-task round trip stays negligible). Each chunk is its own `computeWithArena` call, so between chunks the worker drains its queue and queued renders get through. The pages measured within a budget are accumulated locally and assigned through the `pages` setter once per budget, exactly as before, so the event cadence does not change. Before each chunk the unloaded page list is recomputed on a merged view of the current pages and the pages measured so far in this budget (preserving the `startPageNumber` ordering), and at publish time the measured pages are merged into the current page list rather than a snapshot taken before the first await; a `reloadPages` that completed concurrently is therefore neither overwritten with stale placeholders nor re-measured. `isDisposed` is checked between chunks.

This is complementary to and independent of #708 (priority queue for the worker). With the current FIFO worker, this bounds a render's wait to one chunk plus whatever else was already queued; combined with #708 it bounds the wait to one chunk.

## Behaviour preserved

- Public API unchanged. `onPageLoadProgress` is still invoked once per `loadUnitDuration` budget with the same `(loadedPageCount, totalPageCount, data)` arguments, `PdfDocumentLoadCompleteEvent` semantics are unchanged, and `startPageNumber` ordering is unchanged.
- `_loadPagesInLimitedTime` itself is unchanged apart from a comment; `openFile`'s initial load (`maxPageCountToLoadAdditionally: 1`) is untouched, and `reloadPages` has its own worker task.
- Every chunk still measures at least one page (the timeout is checked after measuring a page), so `Duration.zero` and expired budgets keep making progress exactly as before; the existing `Duration.zero` tests still pass unchanged.
- The page list is still published via the `pages` setter once per `loadUnitDuration` budget, so `PdfDocumentPageStatusChangedEvent` is emitted at the same cadence as before (about 4/s with the default budget) and `PdfViewer` relayouts are not multiplied by the chunking.
- WASM and CoreGraphics backends are not affected (no shared worker there) and are not touched.

## Numbers

Engine-only benchmark: open with `useProgressiveLoading`, measure the target page, then render it at 1200 px width every 100 ms while `loadPagesProgressively(startPageNumber: target)` runs. Two scanned volumes (281 and 242 pages), Windows desktop.

| | 281 pages, before | 281 pages, after | 242 pages, before | 242 pages, after |
| --- | --- | --- | --- | --- |
| Idle render (p50) | 112 ms | 113 ms | 79 ms | 81 ms |
| Render during measurement, p50 / max | 256 / 361 ms | 125 / 179 ms | 227 / 334 ms | 97 / 111 ms |
| Full measurement, idle | 2374 ms | 2433 ms | 2450 ms | 2492 ms |

Render latency during measurement drops to roughly idle plus one chunk. Idle full-measurement time rises by about 2% (one worker round trip per 20 ms chunk, ~130 chunks for these documents).

## Testing

- New tests in `packages/pdfrx_engine/test/pdf_document_test.dart`:
  - `loadPagesProgressively` loads every page of the 40-page asset and reports monotonically non-decreasing `loadedPageCount` ending at `pages.length`, with both a 1 ms and the default `loadUnitDuration`, with a single completion event.
  - A 1 ms `loadUnitDuration` still yields more than one progress callback.
  - Interleaving: on a generated in-memory 20000-page PDF (`buildBlankPdf` in `test/utils.dart`, since the checked-in assets measure in well under one chunk), a `render()` of the already-loaded first page issued while `loadPagesProgressively(loadUnitDuration: 5 s)` runs completes before the load finishes, while pages are still unloaded, and in under a quarter of the load's wall time. The budget is longer than the whole measurement (~500 ms), so the previous implementation runs the entire document as a single worker task and the test fails against it (render 471 ms of a 489 ms load, completing after the load); with chunking the render takes about 20 ms. The test also asserts that only page 1 is loaded when the render completes, since nothing is published before the first budget ends.
  - Once-per-budget publishing: on the same 20000-page document with the default budget, the number of `PdfDocumentPageStatusChangedEvent`s equals the number of `onPageLoadProgress` callbacks (each budget publishes exactly once even though it runs many ~20 ms chunks), every page other than page 1 appears in exactly one event's `changes`, and a single `PdfDocumentLoadCompleteEvent` is emitted. This guards against publishing per chunk, which would relayout the viewer about 40 times a second.
  - Outward order across chunks: with `startPageNumber: 10000` and the callback stopping after the first default budget, the set of loaded pages (excluding page 1) is exactly the head of the outward sequence 10000, 10001, 9999, 10002, 9998, ...; resuming then loads every page. Each of the two new tests runs in about 1 s.
- `dart test` in `packages/pdfrx_engine`: 64 tests, all passed (3 consecutive runs).
- `dart analyze` in `packages/pdfrx_engine` and `flutter analyze` in `packages/pdfrx`: no new issues (the remaining infos pre-exist on master).
- `doc/Progressive-Loading.md` gained a short paragraph describing the chunked measurement, including that with the FIFO worker a render waits for the current chunk plus anything already queued ahead of it.

### DIFF
--- a/doc/Progressive-Loading.md
+++ b/doc/Progressive-Loading.md
@@ -212,7 +212,7 @@ await document.loadPagesProgressively(
 
 The callback is invoked periodically (every `loadUnitDuration`) as pages are loaded. Return `false` from the callback to stop the loading process early.
 
-Internally, each `loadUnitDuration` slice is sent to the PDFium worker as a series of short measurement tasks rather than one long task. All PDFium calls in a process share that worker, so this keeps page rendering responsive while the rest of the document is being measured: a render requested during progressive loading only waits for the current short task to finish instead of the whole slice. The callback cadence is unaffected; it is still invoked once per `loadUnitDuration`.
+Internally, each `loadUnitDuration` slice is sent to the PDFium worker as a series of short measurement tasks rather than one long task. All PDFium calls in a process share that worker, so this keeps page rendering responsive while the rest of the document is being measured: a render requested during progressive loading waits for the current short task (and anything already queued ahead of it) to finish instead of the whole slice. The callback cadence is unaffected; it is still invoked once per `loadUnitDuration`.
 
 The callback receives the number of pages loaded so far (`loadedPageCount`), the total number of pages, and the optional `data` value passed to [`loadPagesProgressively()`](https://pub.dev/documentation/pdfrx_engine/latest/pdfrx_engine/PdfDocument/loadPagesProgressively.html). `loadedPageCount` is a count of loaded pages, not a page index; it does not depend on the order in which the pages are measured.
 
--- a/doc/Progressive-Loading.md
+++ b/doc/Progressive-Loading.md
@@ -212,6 +212,8 @@ await document.loadPagesProgressively(
 
 The callback is invoked periodically (every `loadUnitDuration`) as pages are loaded. Return `false` from the callback to stop the loading process early.
 
+Internally, each `loadUnitDuration` slice is sent to the PDFium worker as a series of short measurement tasks rather than one long task. All PDFium calls in a process share that worker, so this keeps page rendering responsive while the rest of the document is being measured: a render requested during progressive loading only waits for the current short task to finish instead of the whole slice. The callback cadence is unaffected; it is still invoked once per `loadUnitDuration`.
+
 The callback receives the number of pages loaded so far (`loadedPageCount`), the total number of pages, and the optional `data` value passed to [`loadPagesProgressively()`](https://pub.dev/documentation/pdfrx_engine/latest/pdfrx_engine/PdfDocument/loadPagesProgressively.html). `loadedPageCount` is a count of loaded pages, not a page index; it does not depend on the order in which the pages are measured.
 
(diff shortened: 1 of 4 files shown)


palmoni5:
I’d love to hear if there’s any progress on this; this fix is ​​very important for my app.

espresso3389 (approved):

